PR #8, opened 2020-05-09
Add a way to configure repl commands

theHamsta:
Thanks for this awesome plugin. I really appreciate that it's goal is to hackable (and it already).

I was only missing a way to configure the repl commands. So I added a `plugin_config` module (`plugin_confg` to distinguish config for debuggers/debugees). There maybe a better way to organize this.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -12,6 +12,8 @@ local ns_pos = 'dap_pos'
 local Session = {}
 local session = nil
 
+M.repl = repl
+
 --- For extension of language specific debug adapters.
 --
 -- `adapters.<type>` where <type> is specified in a configuration.
@@ -667,11 +669,6 @@ function M.continue()
   else
     session:_step('continue')
   end
-end
-
-
-function M.repl()
-  repl.open()
 end
 
 
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -5,6 +5,16 @@ local win = nil
 local buf = nil
 local session = nil
 
+M.commands = {
+  continue = {'.continue', '.c'},
+  next_ = {'.next', '.n'},
+  into = {'.into'},
+  out = {'.out'},
+  scopes = {'.scopes'},
+  threads = {'.threads'},
+  frames = {'.frames'},
+  exit = {'exit', '.exit'},
+}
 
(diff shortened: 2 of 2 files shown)


theHamsta:
I personally prefer the mapping that I am used to from gdb/pdb:
```vim
    dap.plugin_config.repl_commands = {
        continue = {".continue", "c"},
        next_ = {".next", "n"},
        into = {".into", "s"},
        out = {".out", "r"},
        scopes = {".scopes", "a"},
        threads = {".threads"},
        frames = {".frames", "f"},
        exit = {"exit", ".exit"}
    }
```

theHamsta:
Maybe `ui_config` is a better name?

mfussenegger:
Glad to hear that it's useful to you. I like the idea. Not sure if it is worth introducing a separate config module. 

What do you think of exposing a `commands` table directly within the `repl` module? I think `dap` could additionally expose `repl` by default as well, so that you could do this:

```
 dap.repl.commands = {
        continue = {".continue", "c"},
        next_ = {".next", "n"},
        into = {".into", "s"},
        out = {".out", "r"},
        scopes = {".scopes", "a"},
        threads = {".threads"},
        frames = {".frames", "f"},
        exit = {"exit", ".exit"}
    }
```

Given the early stage of the plugin I think it would also be an option to change the commands altogether. I didn't put much thought into the names for them. Mirroring the names of other tools would certainly make sense. (Though making them configurable is still a good idea I think)

theHamsta:
Problem here is that we can not export the `repl` module directly: there is a function with same name: `dap.repl()`

mfussenegger:
> Problem here is that we can not export the `repl` module directly: there is a function with same name: `dap.repl()`

Maybe `dap.repl()` could be removed in favour of `dap.repl.open()`

theHamsta:
Reconsidering this, I found that we're actually using the dictionary look-up the wrong way. Something like
```
M.commands = {
  ['.c'] = function(session) session:_step('continue') end,
  continue = function(session) session:_step('continue') end,
  ['next'] = function(session) session:_step('next') end,
  ...
}
```
could be even more hackable and simplify the code. Though the current approach is faster for the user to change.

PR #9, opened 2020-05-09
Set buftype `nofile` for repl

theHamsta:
This prevents messages when closing the repl and its contents were not
written to a file.

From doc:
```
      nofile    buffer is not related to a file, will not be written
```


theHamsta:
I am confused. Can a buffer be `nofile` and `prompt` some how?

mfussenegger:
> Can a buffer be nofile and prompt somehow?

I don't think so. But I wonder if it would make sense to have `prompt` imply `nofile` - maybe it would be worth to create an upstream issue in vim or neovim for that. The behavior has also been bugging me.

PR #12, opened 2020-05-09
Add `dap.up` and `dap.down`

theHamsta:
This implements the basic functionality of up/down, i.e. jump to a different frame in stack.

What is not (yet) implemented is to execute stepping commands relative to current frame. One possibility would be to issue multiple 'stepOut' request until the debugger reaches the current frame before performing the actual stepping. Since this is a bit a hack, I omitted it from this PR.

I noticed, that with the current dictionary approach the users must update their config when new repl commands are added. So either the documentation should be changed to recommend something like this 
```lua
dap.repl.commands.continue = {'continue', '.c'}
dap.repl.commands.next_ = {'next '}
```
or we reverse the look-up direction of this dict (see https://github.com/mfussenegger/nvim-dap/pull/8#issuecomment-626174313) or we to a `vim.tbl_merge` with given defaults.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -81,6 +81,14 @@ local function expand_config_variables(option)
   return ret
 end
 
+local function index_of(items, predicate)
+  for i, item in ipairs(items) do
+    if predicate(item) then
+      return i
+    end
+  end
+  return nil
+end
 
 local function handle_adapter(adapter, configuration)
   assert(type(adapter) == 'table', 'adapter must be a table, not' .. vim.inspect(adapter))
@@ -283,29 +291,14 @@ function Session:event_stopped(stopped)
           current_frame = frame
           self.current_frame = frame
         end
-        frames[frame.id] = frame
+        table.insert(frames, frame)
       end
       if not current_frame then
         return
       end
       jump_to_frame(current_frame, stopped.preserveFocusHint)
 
-      self:request('scopes', { frameId = current_frame.id }, function(_, scopes_resp)
-        if not scopes_resp or not scopes_resp.scopes then return end
-
-        current_frame.scopes = {}
-        for _, scope in pairs(scopes_resp.scopes) do
-
-          table.insert(current_frame.scopes, scope)
-          if not scope.expensive then
-            self:request('variables', { variablesReference = scope.variablesReference }, function(_, variables_resp)
-              if not variables_resp then return end
-
-              scope.variables = variables_resp.variables
-            end)
-          end
-        end
-      end)
+      self:_request_scopes(current_frame)
     end)
   end)
 end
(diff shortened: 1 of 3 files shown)


lua/dap/repl.lua
```diff
@@ -85,6 +91,10 @@ function M.execute(text)
         end
       end
     end
+
+    for _, b in pairs(api.nvim_list_bufs()) do
+      M.append(vim.inspect(require'nvim-treesitter.locals'.collect_locals(b)))
+    end
```
mfussenegger, line 97:
```suggestion
```

Looks like this was added by accident.

lua/dap.lua
```diff
@@ -343,6 +328,52 @@ function Session.event_output(_, body)
...
+        end)
+      end
+    end
+  end)
+end
+
+function Session:_frame_delta(delta)
+  if self.stopped_thread_id and self.threads[self.stopped_thread_id].frames then
+    local frames = self.threads[self.stopped_thread_id].frames
+    local current_frame_index = self:_current_frame_index()
```
mfussenegger, line 366: How about something like this:

```
local function index_of(items, predicate)
  for i, item in ipairs(items) do
    if predicate(item) then
      return i
    end
  end
  return nil
end
```

```suggestion
    local current_frame_index = index_of(frames, function(f) return f.id == current end)
    assert(current_frame_index, 'current_frame must exist within frames')
```

The `_current_frame_index` method could then be deleted. It's only used once and a mix of abstract/concrete.
`index_of` would be a more generic concept (that people may be familiar with from other languages standard libraries) - it would also avoid re-retrieving the `frames` again which are already in scope here.

lua/dap.lua
```diff
@@ -343,6 +328,52 @@ function Session.event_output(_, body)
...
+
+    current_frame_index = current_frame_index + delta
+    if 0 < current_frame_index and current_frame_index <= #frames then
+      self.current_frame = frames[current_frame_index]
```
mfussenegger, line 370: Do you think it would make sense to wrap around? Or would that be confusing?
Or maybe print a message when hitting the top-most or bottom-most frame?

lua/dap/repl.lua
```diff
@@ -75,6 +77,10 @@ function M.execute(text)
     session:_step('stepIn')
   elseif vim.tbl_contains(M.commands.out, text) then
     session:_step('stepOut')
+  elseif vim.tbl_contains(M.commands.up, text) then
+    session:_frame_delta(1)
```
mfussenegger, line 81: Do you think it would make sense to print the frames within the range of the current frame - let's say 2 automatically afterwards? 

So you'd get:

```
dap> .up

  a()
  b()
→ c()
  d()
  e()
```

lua/dap.lua
theHamsta: I guess this would be enough feedback:

dap> .up

  a()
  b()
  c()
  d()
 → e()

dap> .up

  a()
  b()
  c()
  d()
 → e()

lua/dap/repl.lua
theHamsta: :+1:

lua/dap.lua
mfussenegger: Yes I agree. Want to include it? Could also be a follow up - I think this PR is good to go

theHamsta: It's included.

lua/dap.lua
```diff
@@ -335,6 +328,39 @@ function Session.event_output(_, body)
...
+function Session:_frame_delta(delta)
+  if self.stopped_thread_id and self.threads[self.stopped_thread_id].frames then
+    local frames = self.threads[self.stopped_thread_id].frames
+    local current_frame_index = index_of(frames, function(i) return i.id == self.current_frame.id end)
+
+    current_frame_index = current_frame_index + delta
+    if 0 < current_frame_index and current_frame_index <= #frames then
+      self.current_frame = frames[current_frame_index]
+
+      jump_to_frame(self.current_frame, true)
```
mfussenegger, line 359: @theHamsta btw did you set the `preserve_focus_hint` to `true` here deliberately? For me `.up` / `.down` turns out to be way more useful if it jumps to the location

theHamsta: No, not deliberately. At that moment, I actually wanted to have `true` for jump to location. But apparently I have done the opposite.

PR #13, opened 2020-05-10
[WIP]: Virtual text

theHamsta:
Currently this PR is still based on the up-down branch. I will rebase and squash everything if everything is ok with this PR.

Add virtual text annotations with the current values of the locals at the spot of their definition. I use nvim-treesitter to get the definition locations and the scopes of the current stack trace. So without nvim-treesitter and a treesitter parser for the given language installed, this change won't be activated.

The whole functionality is controlled by the variable `vim.g.dap_virtual_text` which can have the values `nil`, `false`, `anything else` and `'all frames'`. Virtual text is deactivated by default.

With `vim.g.dap_virtual_text=true`

current_frame

With `vim.g.dap_virtual_text='all frames'`

all_scopes

An additional requirement are scope definitions for the language to be debugged (currently nvim-treesitter has this only for lua and ruby). I can add this later for python. E.g. by placing the following file into `nvim-treesitter/queries/python/locals.scm`:

```scheme
;;; Programm structure
(module) @scope

(class_definition
  body: (block
          (expression_statement
            (assignment
              left: (expression_list
                      (identifier) @definition.associated))))) @scope

; Function with parameters, defines parameters
(function_definition
  name: (identifier)
  parameters: (parameters
                (identifier) @definition.var))

; Function defines function and scope
(function_definition
  name: (identifier) @definition.function) @scope


;;; Loops
(for_statement
  left: (variables
    (identifier) @definition.var)) @scope

(while_statement) @scope


;;; Assignments
(assignment
  left: (expression_list
    (identifier) @definition.var))

; Walrus operator  x := 1
(named_expression
  (identifier) @definition.var)


;;; REFERENCES
(identifier) @reference
```

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -339,7 +339,7 @@ function Session:event_stopped(stopped)
       jump_to_frame(current_frame, stopped.preserveFocusHint)
 
       if vim.g.dap_virtual_text == 'all frames' then
-        virtual_text.clear_virtual_text()
+        virtual_text.clear_virtual_text(current_frame)
         local requested_functions = {}
         for _, f in pairs(frames) do
           -- Ensure to evaluate the same function only once to avoid race conditions
@@ -382,10 +382,10 @@ function Session.event_output(_, body)
 end
 
 function Session:_request_scopes(current_frame)
-  if vim.g.dap_virtual_text ~= 'all frames' then
-    virtual_text.clear_virtual_text()
-  end
   self:request('scopes', { frameId = current_frame.id }, function(_, scopes_resp)
+    if vim.g.dap_virtual_text ~= 'all frames' then
+      virtual_text.clear_virtual_text(current_frame)
+    end
     if not scopes_resp or not scopes_resp.scopes then return end
 
     current_frame.scopes = {}
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -6,6 +6,7 @@ local api = vim.api
 local log = require('dap.log').create_logger('vim-dap.log')
 local ui = require('dap.ui')
 local repl = require('dap.repl')
+local virtual_text = require('dap.virtual_text')
 local M = {}
 local ns_breakpoints = 'dap_breakpoints'
 local ns_pos = 'dap_pos'
@@ -15,6 +16,7 @@ local bp_conditions = {}
 local last_config = nil
 
 M.repl = repl
+M.virtual_text = virtual_text
 
 --- For extension of language specific debug adapters.
 --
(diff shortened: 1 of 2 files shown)


mfussenegger:
This is really cool :)

I'll have a closer look once the other PR is through.

theHamsta:
Apparently, `'all frames'` has the problem that sometimes not all of the scopes requests get an answer. Dunno if this is just an issue with the debugpy. I encountered no problems with `vim.g.dap_virtual_text=true`

theHamsta:
Hi. I had similar concerns like you: vim plugins in general try to be dependency free. But the idea of nvim-treesitter is similar to nvim-lsp: provide configurations for the different parsers. It's true that it's currently not stable. However, neither is Neovim's LSP support.

I tried to make the interface to nvim-treesitter plugin is superthin. Two functions that give you the nodes for variable definitions and for variable scopes. It's not important what magic is performed background. The returned objects are always built-in treesitter nodes. The worst thing that could happen is that the name of the function changes.

In the end, I liked the implementation how it is right now, since is able find robustly the variable locations independently of the code language. If a fine tuning for a specific language is necessary it happens outside of this repo. The parsing is fast and built-in in nvim.

This PR was merely an experiment to see whether it's possible to have this functionality. So there a few possibilities to proceed:

1. I could take maintainership of this module and update it for changes in nvim's treesitter API and nvim-treesitter. You could merge this PR after some time for nvim-treesitter to stabilize.
2. Keep this PR open. People can experiment with when they want to. The small task of nvim-treesitter (making the right queries to the nvim's built-in treesitter) could be done later by the dap-configs repo that you maybe planned in future
3. I could provide this functionality as nvim-dap extension as a separate plug-in. Probably, I need to hook up into a few events of this repo. Probably, being able to register Lua callbacks makes it easier to extend functionality without replying on internals of this repo  than autocmds.
4. I could keep this functionality for in personal configuration. I suppose that for most people it's to much effort to read into the code of this repo and how to use treesitter to build this functionality for themselves

I am happy with either choice! The fact that it's possible prototype an extension like this so quickly shows that repo is well structured and extensible. Thank you for the improvement over vimspector!

theHamsta:
One off-topic comment: gdb and pdb repeat the last input to the REPL on enter if nothing is entered (stepping for lazy people). Should I make a PR for that behavior?

theHamsta:
Or another idea would be that this repo sets the virtual text but just based on textual occurrences of the locals (like vscode does, I suppose). Users could then overwrite the function that resolves occurrences by something better if they want to.

theHamsta:
Queries for Python have been merged (https://github.com/nvim-treesitter/nvim-treesitter/pull/56). Queries for C/C++ and Rust are on the way (https://github.com/nvim-treesitter/nvim-treesitter/pull/57, https://github.com/nvim-treesitter/nvim-treesitter/pull/63).

image

mfussenegger:
If I understand  https://github.com/neovim/neovim/pull/12639 right, it will allow us to build this without any direct nvim-treesitter dependency at all. Right?

vigoux:
Hi, not sure... Because you might need the `locals.scm` queries anyway. I really think that they will be upstreamed in the end, but for now I quite advise still using `nvim-treesitter`.

theHamsta:
nvim-treesitter has got a more functionality in resolving definitions so I code from this PR could be removed. Also a lot more languages are supported now.

theHamsta:
Is now available as a separate plugin after merging #37

PR #15, opened 2020-05-12
Add Goto command

theHamsta:
- Add a `.goto` repl command. E.g. `.goto 10` to jump to line 10. This is supported by debugpy. Debugpy ignores all other arguments.

- Add `dap.goto_({line}`) to jump to line under cursor or a specific line.

In theory, the debug server could offer more than one jump target. I am just taking the first one if any. 

## GotoTargets Request

This request retrieves the possible goto targets for the specified source location.

These targets can be used in the ‘goto’ request.

Clients should only call this request if the capability ‘supportsGotoTargetsRequest’ is true.
```
interface GotoTargetsRequest extends Request {
  command: 'gotoTargets';

  arguments: GotoTargetsArguments;
}
```

Arguments for ‘gotoTargets’ request.

```
interface GotoTargetsArguments {
  /**
   * The source location for which the goto targets are determined.
   */
  source: Source;

  /**
   * The line location for which the goto targets are determined.
   */
  line: number;

  /**
   * An optional column location for which the goto targets are determined.
   */
  column?: number;
}
```

Response to ‘gotoTargets’ request.

```
interface GotoTargetsResponse extends Response {
  body: {
    /**
     * The possible goto targets of the specified location.
     */
    targets: GotoTarget[];
  };
}
```

## GotoRequest

```
interface GotoRequest extends Request {
  command: 'goto';

  arguments: GotoArguments;
}
```

Arguments for ‘goto’ request.
```
interface GotoArguments {
  /**
   * Set the goto target for this thread.
   */
  threadId: number;

  /**
   * The location where the debuggee will continue to run.
   */
  targetId: number;
}
```

Response to ‘goto’ request. This is just an acknowledgement, so no body field is required.

```
interface GotoResponse extends Response {
}
```

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -357,6 +357,35 @@ function Session:_request_scopes(current_frame)
   end)
 end
 
+--- Goto specified line (source and col are optional)
+function Session:_goto(line, source, col)
+  local frame = self.current_frame
+  if not frame then
+    return
+  end
+  if not self.capabilities.supportsGotoTargetsRequest then
+    print("Debug Adapter doesn't support GotoTargetRequest")
+    return
+  end
+  self:request('gotoTargets',  {source = source or frame.source, line = line, col = col}, function(err, response)
+    if err then
+      print('Error getting gotoTargets: ' .. err.message)
+      return
+    end
+    if not response or not response.targets then
+      print("No goto targets available. Can't execute goto")
+      return
+    end
+    local params = {threadId = session.stopped_thread_id, targetId = response.targets[1].id }
+    self:request('goto', params, function(err1, _)
+      if err1 then
+        print('Error executing goto: ' .. err1.message)
+      end
+    end)
+  end)
+
+end
+
 function Session:_frame_delta(delta)
   if not self.stopped_thread_id then
     print('Cannot move frame if not stopped')
@@ -711,6 +740,18 @@ end
 function M.down()
   if session then
     session:_frame_delta(-1)
+  end
+end
+
+function M.goto_(line)
+  if session then
+    local source, col
+    if not line then
+      line, col = unpack(api.nvim_win_get_cursor(0))
+      col = col + 1
+      source = { path = vim.uri_from_bufnr(0) }
+    end
+    session:_goto(line, source, col)
   end
 end
 
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 - [ ] set exception breakpoints
 - [x] step over, step into, step out
 - [ ] step back, reverse continue
-- [ ] Goto
+- [x] Goto
 - [x] restart
 - [x] stop
 - [x] evaluate expressions
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -207,6 +207,13 @@ down()                                                              *dap.down()*
         Go down in current stacktrace without stepping.
 
 
+goto_({line})                                                      *dap.goto_()*
+        Let the debugger jump to a specific line or line under cursor.
+
+        Parameters: ~
+            {line}  Line number or line under cursor if nil.
+
+
 repl.open()                                                    *dap.repl.open()*
         Open a REPL / Debug-console.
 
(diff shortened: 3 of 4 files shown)


lua/dap.lua
```diff
@@ -347,6 +347,27 @@ function Session:_request_scopes(current_frame)
...
+          self:request('goto', {threadId = session.stopped_thread_id, targetId = response.targets[1].id },
+          function(err, _)
+            if err then
+              print(err.message)
+            end
+          end)
+        end
+      end
+    end)
+  end
```
mfussenegger, line 368: I think it would help with the readability to add some early returns, to reduce the number of indentation levels a bit.

And it might make sense to check for `supportsGotoTargetsRequest` and give the user a message if it's not supported.

```suggestion
  local frame = self.current_frame
  if not frame then
    return
  end
  if not self.capabilities.supportsGotoTargetsRequest then
    print('Debug Adapter doesn't support GotoTargetRequest')
    return
  end
  self:request('gotoTargets',  {source = source or frame.source, line = line, col = col}, function(err, response)
    if err then
      print('Error getting gotoTargets: ' .. err.message)
      return
    end
    if not response or not response.targets then
      print('No goto targets available. Can't execute goto')
      return
    end
    local params = {threadId = session.stopped_thread_id, targetId = response.targets[1].id }
    self:request('goto', params, function(err, _)
      if err then
        print('Error executing goto: ' .. err.message)
      end
    end)
  end)

```

lua/dap.lua
```diff
@@ -678,6 +699,17 @@ function M.down()
   end
 end
 
+function M.goto_(line)
+  if session then
+    local source, col
+    if not line then
```
mfussenegger, line 705: Shouldn't the `source` also be set if the `line` is set?

lua/dap.lua
```diff
@@ -678,6 +699,17 @@ function M.down()
...
+      line, col = unpack(api.nvim_win_get_cursor(0))
+      source = { path = vim.uri_from_fname(vim.fn.expand('%:p')) }
```
mfussenegger, line 707:
```suggestion
      source = { path = vim.uri_from_bufnr(0) }
```

lua/dap.lua
```diff
@@ -347,6 +347,27 @@ function Session:_request_scopes(current_frame)
   end)
 end
 
+function Session:_goto(line, source, col)
```
mfussenegger, line 350: Doesn't matter much, but I think one of the following parameter ordering variants would make more sense.

```suggestion
function Session:_goto(line, col, source)
```

(source last because it's optional)



```suggestion
function Session:_goto(source, line, col)
```

(source -> line -> col, because line is inside the source, and col is inside the line)

lua/dap.lua
```diff
@@ -678,6 +699,17 @@ function M.down()
   end
 end
 
+function M.goto_(line)
+  if session then
+    local source, col
+    if not line then
+      line, col = unpack(api.nvim_win_get_cursor(0))
```
mfussenegger, line 706: `col` here is 0-indexed. Is that also what the DAP expects for the gotoTargetsRequest?

theHamsta: No. in the init request. You specified that column start at 1. I'llf fix that

lua/dap.lua
theHamsta: Actually, only the line important. This is the only information that is processed by real life adpaters. Everything else is optional-

lua/dap.lua
theHamsta: Source is optional, since it almost always given by the current stopped frame.

lua/dap.lua
mfussenegger: > . I tried to order by how probable I estimate that the given information is ever necessary. DAP protocol expects line and source. col is optional

Ok makes sense.  Let's keep it then. 

But I think if col is optional then the col + 1 can fail

theHamsta: I put it right after the get_cursor call.

PR #17, opened 2020-05-22
Reset arrow after frame switch

theHamsta:
~~I guess you renamed: `_set_frame` to `_frame_set`.~~

~~This PR fixes up/down for that change.~~

Resets the `->` after pressing enter on a frame. ~~If it's to much complexity for this mini-feature I'll just drop the corresponding commit.~~

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -98,6 +98,19 @@ function M.on_enter()
     local frame = frames_marks[mark_id]
     if frame then
       session:_frame_set(frame)
+      local new_marks = {}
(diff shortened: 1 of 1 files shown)


theHamsta:
Apparently, there's still a bug in the code. Setting this to WIP and made a separate PR for the bugfix.

theHamsta:
Hm, also without this commit applied enter some times does not change the frame on some line when I switch the frame often.

theHamsta:
Hmm, the bug seems to occur also without this commit applied. When I switch the frames really often sometimes enter does not change the frame.

PR #18, opened 2020-05-22
Fix up/down: _set_frame -> _frame_set

theHamsta:
I guess you renamed: `_set_frame` to `_frame_set`.

This PR fixes up/down for that change.


mfussenegger:
Yep I missed that :(

Thank you for the quick fix

PR #19, opened 2020-05-24
Add support for reverse debug

theHamsta:
Reading the DAP docs, I guess `stepBack` and `reverseContinue` just work as the normal stepping commands.

Didn't find the time yet to test with `lldb-vscode`+`rr` or with https://github.com/Microsoft/vscode-mock-debug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ things may break.
 - [ ] set function breakpoints
 - [x] set exception breakpoints
 - [x] step over, step into, step out
-- [ ] step back, reverse continue
+- [x] step back, reverse continue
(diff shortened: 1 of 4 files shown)


mfussenegger:
@theHamsta do you know of any debug adapter that supports this?

theHamsta:
`rr` and the demo debug adapter that can "debug" on markdown. The demo debug adapter supports all features.

theHamsta:
Tested this with https://github.com/Microsoft/vscode-mock-debug and added instructions for mock-debug to the wiki. Works as expected.

PR #21, opened 2020-06-07
Fix variable to omnifunc

ayyess:
I was getting an error on C-xC-o and it looked like the omnifunc definition was the odd one out in the file.

```
E5108: Error executing lua [string "v:lua"]:1: attempt to call field 'omnifunc' (a nil value)
```

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -1,6 +1,3 @@
-dap = {} -- luacheck: ignore 111 - to support v:lua.dap... uses
-
-
 local uv = vim.loop
 local api = vim.api
 local log = require('dap.log').create_logger('vim-dap.log')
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thank you for the PR.

Do you by chance have a `dap = require('dap')` in your configuration?
I think that could be overriding the `dap` definition that this plugin is doing by itself.

The change would probably break the functionality for users who have a different import (`local dap = require..`).
Maybe it could be changed to keep the `dap` definition, but do something like this:


```
local function omnifunc...

M.omnifunc = omnifunc
dap.omnifunc = omnifunc
```

ayyess:
I see now that `dap` is global and that I was misunderstanding lua's scope. Thanks for explaining.

mfussenegger:
> I see now that `dap` is global and that I was misunderstanding lua's scope. Thanks for explaining.

So this change is not required then, right?

ayyess:
Right

PR #36, opened 2020-08-23
Add support for exception breakpoints

theHamsta:
```lua
-- Ask user to stop on which kinds of exceptions
require'dap'.set_exception_breakpoints()
-- don't stop on exceptions
require'dap'.set_exception_breakpoints({})
-- stop only on uncaughted exceptions
require'dap'.set_exception_breakpoints({"uncaughted"})
-- stop on all exceptions
require'dap'.set_exception_breakpoints({"raised", "uncaught"})

```

TODO:

 - [x] docs

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -281,7 +281,7 @@ function Session:event_initialized(_)
 
   self:set_breakpoints(nil, function()
     if self.capabilities.exceptionBreakpointFilters then
-      self:set_exception_breakpoints('default', on_done)
+      self:set_exception_breakpoints('default', nil, on_done)
     else
       on_done()
     end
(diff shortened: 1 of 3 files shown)


lua/dap.lua
```diff
@@ -333,6 +410,9 @@ function Session:event_stopped(stopped)
     return
   end
   self.stopped_thread_id = stopped.threadId
+  if stopped.reason == 'exception' then
+    self:_show_exception_info()
+  end
```
mfussenegger, line 415: Do you think we should chain the `threads`  request call below after the `exceptionInfo` response has been processed? 

I think otherwise the way the repl buffer is mutated may not have a fixed order.

lua/dap.lua
```diff
@@ -325,6 +333,75 @@ local function jump_to_frame(frame, preserve_focus_hint)
   end
 end
 
+--interface ExceptionInfoRequest extends Request {
```
mfussenegger, line 336: So far we didn't add the request structure definitions, not sure if we should start doing that now - and if so, maybe closer to the structure definitions where we use them?

theHamsta: We could also replace it by a link to the specification.

lua/dap.lua
theHamsta: :+1: It has to be the other way around. `exceptionInfo` needs to know `threadId`.

lua/dap.lua
theHamsta: Added a short documentation with links to specification at the points we use the request/responses

mfussenegger (approved):

PR #37, opened 2020-08-23
Add support for user-defined `custom_event_handlers`

theHamsta:
This enables extensions to nvim-dap both defined either by end users or
by plugin developers. This enables me to implement
https://github.com/mfussenegger/nvim-dap/pull/13 as an external plugin.

Usage to hook into `event_continued`:

```lua
if dap.custom_event_handlers then
    table.insert(dap.custom_event_handlers.event_continued, function(session, body)
      dap.repl.append(vim.inspect(body))
    end)
end
```
or
```lua
dap.custom_event_handlers.event_exited["my handler id"] = function(_, _)
   dap.repl.close()
end
```

If this PR has your general ok. I could implement the external plugin to see if these changes are reasonable.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -15,6 +15,18 @@ local bp_conditions = {}
 local last_run = nil
 
 M.repl = repl
+M.custom_event_handlers = setmetatable({}, {
+  __index = function(tbl, key)
+    rawset(tbl, key, {})
+    return rawget(tbl, key)
+  end
+})
+M.custom_response_handlers = setmetatable({}, {
+  __index = function(tbl, key)
+    rawset(tbl, key, {})
+    return rawget(tbl, key)
+  end
+})
 
 --- For extension of language specific debug adapters.
 --
@@ -536,6 +548,9 @@ function Session:handle_body(body)
     self.message_callbacks[decoded.request_seq] = nil
     if decoded.success then
       callback(nil, decoded.body)
+      for _, c in pairs(M.custom_response_handlers[decoded.command]) do
+        c(self, decoded.body)
+      end
     else
       callback({ message = decoded.message; body = decoded.body; }, nil)
     end
@@ -543,6 +558,9 @@ function Session:handle_body(body)
     local callback = self['event_' .. decoded.event]
     if callback then
       callback(self, decoded.body)
+      for _, c in pairs(M.custom_event_handlers['event_' .. decoded.event]) do
+        c(self, decoded.body)
+      end
     else
       log.warn('No event handler for ', decoded)
     end


lua/dap.lua
```diff
@@ -15,6 +15,18 @@ local bp_conditions = {}
...
+    rawset(tbl, key, {})
+    return rawget(tbl, key)
+  end
+})
+M.custom_response_handlers = setmetatable({}, {
+  __index = function(tbl, key)
+    rawset(tbl, key, {})
+    return rawget(tbl, key)
+  end
+})
```
theHamsta, line 29: Could also be one table. Don't know whether it's reasonable to separate between events and request responses.

lua/dap.lua
```diff
@@ -536,13 +548,19 @@ function Session:handle_body(body)
     self.message_callbacks[decoded.request_seq] = nil
     if decoded.success then
       callback(nil, decoded.body)
+      for _, c in ipairs(M.custom_response_handlers[decoded.command]) do
+        c(self, decoded.body)
+      end
```
theHamsta, line 553: This is mainly to update information from session on requests that add additional information (e.g. requested scopes). So an UI can update whenever new scopes arrived. UI may also do their own requests, but this may lead to duplication of the requests.

mfussenegger:
> If this PR has your general ok

Yes look good to me :+1:

lua/dap.lua
mfussenegger: I'm also not sure, but given that the spec distinguishes between events and responses I think it doesn't hurt to make the distinction explicit here too. Could help avoid ambiguity in the future. So :+1: for separating it.

lua/dap.lua
mfussenegger: > A alternative would be to just have a session_information_updated event, which would UIs cause to update their information from the session instead of right after specific info requests.

Sounds also interesting, though I think that could even be done additionally - so that both lower level hooks and higher level hooks are present.

Maybe you could try out the https://github.com/mfussenegger/nvim-dap/pull/13 use case to see what works better?

theHamsta:
Worked pretty well: https://github.com/theHamsta/nvim-dap-virtual-text/blob/master/lua/nvim-dap-virtual-text.lua

- clear virtual text on `continue`, `exited`
- set virtual text on `variables` response

If I want variables for all frames, I can request them on `stackTrace` response.

mfussenegger (approved):

lua/dap.lua
theHamsta: After using the new plugin for a while I observed that my that my response handlers are not always executed after the built-in ones. Maybe high-level events like session_variables_changed/session_changed might help? Or is there a way that I can be sure that nvim-dap has already handled the responses? Will `callback` only schedule the callback instead of running it immediately?

https://github.com/theHamsta/nvim-dap-virtual-text/issues/1

mfussenegger: Could this be caused be callbacks that in turn invoke other requests?


With a call-stack like this?

```
req1
resp1
  callback
    req2
    resp2?
    custom_callback_for_resp_2?
  custom_callback_for_resp1
```

I'm not sure if that's what is happening.

Would it be an option to move `callback(nil, decoded.body)` after the custom response handlers? To try out if this is the issue.

PR #38, opened 2020-08-24
Logpoints

theHamsta:
Add support for logpoints and listing of breakpoints.

TODO:
- [x] docs

Log points are essentially printf debugging. You can use variables in your log message: `value of foo is {foo}`.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -28,6 +28,9 @@ M.custom_response_handlers = setmetatable({}, {
   end
 })
 
+local DAP_QUICKFIX_TITLE = "DAP Breakpoints"
+local DAP_QUICKFIX_CONTEXT = DAP_QUICKFIX_TITLE
+
 --- For extension of language specific debug adapters.
 --
 -- `adapters.<type>` where <type> is specified in a configuration.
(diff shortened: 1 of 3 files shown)


lua/dap.lua
```diff
@@ -845,24 +860,68 @@ function M.restart()
...
+  local num_bufs = #bp_signs
+  local qf_list = {}
+  for _, buf_bp_signs in pairs(bp_signs) do
+    local bufnr = buf_bp_signs.bufnr
+    for _, bp in pairs(buf_bp_signs.signs) do
+      local condition = bp_info[bp.id].condition;
+      local hitCondition = bp_info[bp.id].hitCondition;
+      local logMessage = bp_info[bp.id].logMessage;
+      local text = table.concat(vim.tbl_filter(function(t) return t end, {
+               unpack(api.nvim_buf_get_lines(bufnr, bp.lnum - 1, bp.lnum, false), 1),
```
mfussenegger, line 874: I may be missing something,  but it looks like the tbl_filter call here is a noop and could be removed?

mfussenegger (approved):
Really cool,  I wanted add this stuff as well but hadn't got around to doing it.  Thanks a lot

lua/dap.lua
theHamsta: It filters out `nil` from the table, otherwise I remember getting an error that `nil` can not be concatenated.

theHamsta: You were right. Maybe the error I remember was caused by something different. If we ever observe an error here we can still add the filtering again.

In this case, `nil` shouldn't cause any problems. But it still did: https://github.com/nvim-treesitter/nvim-treesitter/pull/341#event-3689877983

lua/dap.lua
```diff
@@ -848,24 +863,68 @@ function M.restart()
...
+        condition = condition,
+        logMessage = log_message,
+        hitCondition = hit_condition
+      }
     end
   end
   if session and session.initialized then
     session:set_breakpoints(bufnr)
   end
+  M.list_breakpoints(false)
```
mfussenegger, line 927: Not sure if I like `toggle_breakpoint` to automatically update the quickfix list and potentially override other things. I think if a user likes this behavior, it would be easy enough to create a mapping that calls both functions.

Would you be okay with removing this?

theHamsta: Sure, I was also not sure how to handle this. But at least we are now overwriting other quick fixes when using our own quick fix context. It could be that we set the focus to our quick fix since it's the newest. We could only do it if our quick fix is already populated.

But I'm also fine with removing. Users can also update it after the setBreakpoints response if they want to.

mfussenegger: > We could only do it if our quick fix is already populated.

You mean if a user triggered `list_breakpoints` and the current quickfix window contains breakpoints, toggling a breakpoint would update the list. But if the current quickfix window contains anything else it wouldn't update the list?

That sounds also good to me :+1:

theHamsta: The changed version will only update the qflist if it's the current one. It will however not update the quickfix list if it's been replaced by a newer quickfix list by another plugin.

```
:help getqflist-examples
:help colder
```

theHamsta: Should we auto-close when our quickfix list is empty? I guess it's better to leave this to the user.

mfussenegger: > Should we auto-close when our quickfix list is empty? I guess it's better to leave this to the user.

Agreed, I'd leave that to the user as well.

PR #41, opened 2020-08-26
Handle request after close race condition

mfussenegger:


theHamsta:
You could also avoid setting the `message_call` ever to `nil` an thus leaving it in a undefined state when there's not session (ony reset on session start). But this shouldn't make any big difference.

mfussenegger:
> You could also avoid setting the `message_call` ever to `nil` an thus leaving it in a undefined state when there's not session (ony reset on session start). But this shouldn't make any big difference.

Good point, changed it to reset the `message_callbacks` to `{}`

Issue #6, opened 2020-04-19
nvim-dap News/Updates (Please Subscribe)


mfussenegger:
To get notifications about (major) updates to nvim-dap subscribe to this issue.

mfussenegger:
It is now possible to jump to a frame within the REPL by pressing `enter` on an entry after listing the frames with `.frames`:

![screenrecord](https://user-images.githubusercontent.com/38700/82694240-34d08e80-9c63-11ea-8794-8afab207d04c.gif)

mfussenegger:
There's a new extension called [nvim-dap-python](https://github.com/mfussenegger/nvim-dap-python/) which makes debugging python tests a bit easier:

![asciinema-e9d4f431c04901f3asciicast](https://user-images.githubusercontent.com/38700/100256365-ffc03400-2f44-11eb-93c3-9bc8fecb4953.gif)

mfussenegger:
Thanks to @theHamsta there is now a variable floating view:


https://user-images.githubusercontent.com/38700/105576420-fe5b8e00-5d72-11eb-8e4e-3063506597a1.mp4

mfussenegger:
Merged a change that makes `${file}` always resolve to an absolute path. If you used `${workspaceFolder}/${file}` in your configurations before you have to change it. See https://github.com/mfussenegger/nvim-dap/pull/353 for more context.

mfussenegger:
It is now possible to change the value of variables via either a "Set expression" or "Set Variable" action in the scopes widget. Which action is available depends on what the debug adapter supports:


https://user-images.githubusercontent.com/38700/144464930-e41c6258-9efe-4306-bd25-6d3d5ee1cc98.mp4


(The action menu is bound to `a` per default)

mfussenegger:
Closing this, instead there will be releases with release notes in the future

PR #44, opened 2020-08-27
Variable UI

theHamsta:
Very first draft for a variable UI. I know that nvim-dap wants to be minimal so maybe this could be another extension. *This is only a first try how one could do it*. Let's just think a few week about how to handle this or whether to handle it at all.

I tried to place a expandable tree on a buffer. So it could be used as a hover or as a watch window or however users want to use it. I don't know why LSP hover is not focusable at the moment. I think there's an upstream PR for it. 

Right now there's now keymaps for interaction set and no highlighting. No update after DAP events like `continue`. No highlighting.

screenshot
image


theHamsta:
It would be handy to set some keys like `<enter>` to expand more levels:
image

mfussenegger (commented):
I think the plugin could certainly use a better UI. Especially inspecting complex structures is currently a bit troublesome.

I kinda like the approach taken in https://github.com/norcalli/snippets.nvim/tree/master/lua/snippets/inserters that there are different UIs for the snippet expansion.

I wonder if we could follow a similar approach here - that there could be different UI variants and users could use the ones that best fit.

I think that would imply that `lua/dap.lua` would need to stay fairly de-coupled and be unaware of any concrete UI implementation. So the UIs would use the current API or potentially the hooks you recently added. 

The different ui variants could maybe go into a `uis` folder?

theHamsta:
For this UI I needed almost no change for `dap.lua` (except providing a function for calling it). We can have completely decoupled UIs. UI can subscribe to DAP events to know when the state of the session changed.

We only have to agree how additional information should be inserted into `session`'s data structures. Since it's kind of a tree I just inserted the requested variables into its parent. Then UIs only need to request information that's not already there.

theHamsta:
https://github.com/neovim/neovim/pull/12720

theHamsta:
The variable-tree is not expandable. I still need to figure out how to dispatch on different UI approaches. There's two possibilities:

- provide a generic `inspect variable(s)`: each UI can handle this differently

or

- provide different functions to open buffers for inspecting a set of variables (e.g. hover for variable under cursor, sidebar for locals)
-  provide a generic `inspect variable(s)` that operates on a buffer: each UI can fill their buffer differently

Right now I have taken the second approach. `variable-tree` just takes `buffer, window, session, variables` (window is optional if window size should be adapted) which is ok. But the interface for creating the variable windows is still rubbish.

lua/dap.lua
```diff
@@ -75,6 +78,7 @@ M.adapters = {}
 --
 M.configurations = {}
 
+function M.session() return session end
```
theHamsta, line 81: When we want to have no change to `dap.lua` UIs need somehow access to the session.

theHamsta:
Example usage:
```
    nmap <buffer> <silent> <leader>dh :lua require 'dap.ui.variable-tree'.hover(require 'dap'.session())<cr>
    nmap <buffer> <silent> <leader>dl :lua require 'dap.ui.variable-tree'.open_sidebar(require 'dap'.session())<cr>
```

lua/dap/ui/variable-tree.lua
```diff
@@ -0,0 +1,246 @@
...
+  end
+end
+
+function M.open_sidebar(session, steal_focus, winopts, variables)
+  if not session then
+    return
+  end
+  if not session.stopped_thread_id then
+    return
+  end
```
mfussenegger, line 239: I'm not sure about having the session as argument here.

I'd expect to be able to keep the side bar open across debug sessions and that it would get updated on a new session.

Could the sidebar register a `custom_event_handlers` for `event_initialized` to get access to the session and unsubscribe if the buffer gets detached?

lua/dap/utils.lua
```diff
@@ -0,0 +1,68 @@
...
+  local rtn = {}
+  local callable = iscallable(fn)
+  for _, v in pairs(tbl) do
+    rtn[callable and fn(v) or fn] = v
+  end
+  return rtn
+end
+
+
+function M.map_to_value(fn, tbl)
```
mfussenegger, line 26: Looks like this is currently unused?

lua/dap/utils.lua
```diff
@@ -0,0 +1,68 @@
...
+    return type(mt) == "table" and type(mt.__call) == "function"
+  else
+    return false
+  end
+end
+
+
+function M.map_to_key(fn, tbl)
+  local rtn = {}
+  local callable = iscallable(fn)
```
mfussenegger, line 18: Looks like currently all call-sites always pass in a function. I think the `iscallable` could be removed for now and we can have the contract that `fn` needs to be a function.

lua/dap/utils.lua
```diff
@@ -0,0 +1,68 @@
...
+  elseif type(x) == 'table' then
+    local mt = debug.getmetatable(x)
+    return type(mt) == "table" and type(mt.__call) == "function"
+  else
+    return false
+  end
+end
+
+
+function M.map_to_key(fn, tbl)
```
mfussenegger, line 16:
```suggestion
function M.group_values(values, get_key)
```

Took me a bit to get what the function does. To me `map` feels a bit overloaded in this context because it is often used to refer to the type `Map` (a dictionary, mapping key to values), but also in the context of "mapping a function over values"

lua/dap/ui/variable-tree.lua
```diff
@@ -0,0 +1,246 @@
...
+  return line, max_textlength
+end
+
+function M.hover(session)
+  if not session then
+    return
+  end
+  if not session.stopped_thread_id then
+    return
+  end
```
mfussenegger, line 209: Given that `M.hover` is likely called explicit with a mapping by the user, I think here it would make sense to give some feedback if there is no session or no stopped thread

lua/dap/ui/variable-tree.lua
```diff
@@ -0,0 +1,246 @@
...
+--- name: string
+--- value: string (since we're also accepting scopes here: also optional)
+--- type?: string
+--- presentationHint?: VariablePresentationHint
+--- evaluateName?: string
+--- variablesReference: number
+--- namedVariables?: number
+--- indexedVariables?: number
+--- memoryReference?: string
+function M.write_variables(buf, variables, line, column, win)
```
mfussenegger, line 145: Do you  intend for users to call this directly?

mfussenegger (commented):
Left a couple of comments
The functionality is nice :+1:

lua/dap/ui/variable-tree.lua
theHamsta: It's a remain from the time it was a in a more low-level module that would only write in existing buffers.

lua/dap/ui/variable-tree.lua
```diff
@@ -0,0 +1,246 @@
...
+  M.write_variables(buf, state.variables, 0, 0, win)
+end
+
+function M.toggle_variable_expanded()
+  local buf = api.nvim_get_current_buf()
+  local state = variable_buffers[buf]
+  local pos = api.nvim_win_get_cursor(0)
+  local win = api.nvim_get_current_win()
+
+  local v = state.line_to_variable[pos[1] - 1]
```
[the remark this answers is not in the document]
theHamsta: It should not be possible. The autocmd and the state are created in successive line. It could happen if the user calls this internal function. I added a check.

lua/dap/ui/variable-tree.lua
theHamsta: Yes, I planned this as a future extension. The buffer could subscribe to certain events when it should update variables. I will have a look into how to do this.

I could subscribe to `event_subscribe` to create a global sidebar that always shows the current scopes. How it's written right now, a window is opened to inspect a certain set of variables. So you can have a window for certain scopes of different frames, or for a certain set variables that you want to watch.

I could write some functions actually using the functionality and subscribing to session protocols to implement a local scopes or a watch window.

lua/dap/utils.lua
theHamsta: renamed to `calc_kv_table_from_values`

lua/dap/ui/variable-tree.lua
theHamsta: Added a message.

theHamsta:
@mfussenegger 

A short update on this. I decided to remove the sidebar functionality for now since it might require some update logic. Can we just merge the hover window as soon as https://github.com/neovim/neovim/pull/12720 gets merged?

numpy
Without the mentioned PR you cannot enter the floating window and expand the variables.

mfussenegger:
> Can we just merge the hover window funcionality as soon as neovim/neovim#12720 gets merged?

Yes.  I actually prefer getting smaller changes in one at a time 👍 
I'll test this for a while and get it merged once the neovim PR is merged

mfussenegger:
Finally merged this with some additions made in https://github.com/mfussenegger/nvim-dap/pull/88
I made some (maybe opinionated) changes, I hope you don't mind.

Thank you for the nice work on this :+1:

theHamsta:
Thanks for making this ready! I guess it would be even nicer when we can actually focus the hover window and expand the variables.

mfussenegger:
> Thanks for making this ready! I guess it would be even nicer when we can actually focus the hover window and expand the variables.

This should now work. See the demo https://github.com/mfussenegger/nvim-dap/issues/6#issuecomment-765905896  and the changes here https://github.com/mfussenegger/nvim-dap/pull/88/commits/2e9600eb8ebb70863492a4db1fd3cf92b7518449

As a side effect the floating window doesn't close automatically anymore. I wasn't sure how to best make it focusable while keeping that behavior.

theHamsta:
Well, best would be to have the LSP windows focusable in upstream again (originally it was focusable). But I guess we need to wait a bit more for that to happen.

theHamsta:
When using this feature in the past month I also thought that it would be nice to have the resolve function for the hover configurable. At the moment the expression for the hover is `vim.fn.eval('<cword>')`. But I often wanted to also evaluate a function `foo('bar')`. With tree-sitter one could use a query files to capture possible expressions that can be evaluated. Maybe also just try the current node as an expression.

mfussenegger:
> Well, best would be to have the LSP windows focusable in upstream again (originally it was focusable). But I guess we need to wait a bit more for that to happen.

Yes, I think there was some boarder discussion about a popup API. Maybe taking inspiration from [plenary](https://github.com/nvim-lua/plenary.nvim) or [popup.nvim](https://github.com/nvim-lua/popup.nvim). We'll see.


>  also thought that it would be nice to have the resolve function for the hover configurable. At the moment the expression for the hover is vim.fn.eval('<cword>'). But I often wanted to also evaluate a function foo('bar')

I guess  we could expose something like a `M.resolve_variable_name` which defaults to `vim.fn.eval('<cword>`)` and which users could override. Or another approach would be to have a `visual_hover()` and use the selection.

PR #46, opened 2020-08-29
Add linter github action

mfussenegger:


theHamsta:
Might be useful: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/.luacheckrc

mfussenegger:
> Might be useful: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/.luacheckrc

Yeah there was already a luacheck config in the repo, but the linter action required it to be under a diifferent path

PR #50, opened 2020-09-14
Expand variables with structure after evaluate in REPL

mfussenegger:
Maybe this should also allow expansion of further nested attributes via `<CR>`, similar to how it works on the frames?

asciinema-1ad14b9d7b9b4426asciicast

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -56,6 +56,35 @@ function M.print_stackframes()
 end
 
 
+local function evaluate_input(text, lnum)
+  session:evaluate(text, function(err, resp)
+    if err then
+      M.append(err.message, lnum)
+      return
+    end
+    if resp.variablesReference == 0 then
+      M.append(resp.result, lnum)
+      return
+    end
+
+    local params = {
+      variablesReference = resp.variablesReference
+    }
+    session:request('variables', params, function(err1, v_resp)
+      if err1 then
+        M.append(err1.message, lnum)
+        return
+      end
+
+      M.append(resp.result, lnum)
+      for _, val in ipairs(v_resp.variables) do
+        M.append('  ' .. val.name .. ': ' .. val.value)
+      end
+    end)
+  end)
+end
+
+
 local function execute(text)
   if text == '' then
     if history.last then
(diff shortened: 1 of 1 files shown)


theHamsta:
Could could have a look how I did it in the variable UI. It's a tree structure that can draw on a buffer and expand/unexpand the nested structures on <enter>.

I save the evaluated variables in as a subfield `variables` of each variable. It's handy to have the variables not as a list but as a dictionary because look-up for values is much easier `variables['foo']`.

mfussenegger:
> You could could have a look how I did it in the variable UI. It's a tree structure that can draw on a buffer and expand/unexpand the nested structures on `<enter>`.
> 
> I save the evaluated variables in as a subfield `variables` of each variable (the root are the scopes of each frame). It's handy to have the variables not as a list but as a dictionary because look-up for values is much easier by name `variables['foo']`.

I'll have a closer look, thanks for the pointer. :+1:

theHamsta:
The only important file is `variable-tree.lua` should be relevant. The rest is just some random changes...

theHamsta:
But one experience I made: you have to sort the variables after expansion. It's a lot easier to find things and indexed variables get printed in the correct order.

theHamsta:
Regarding the hover/watch window from #44, do you think this should be a separate extension or do you want to create an API for variable inspection that can show variables in different places repl/float window/split with a general tree implementation in core?

I realized that just having `variable-tree.lua` you need to change nothing in core: Just a command/function that triggers the hover

mfussenegger:
> Regarding the hover/watch window from #44, do you think this should be a separate extension or do you want to create an API for variable inspection that can show variables in different places repl/float window/split with a general tree implementation in core?
> 
> I realized that just having `variable-tree.lua` you need to change nothing in core: Just a command/function that triggers the hover

At the current state I'd like to avoid trying to create a generic abstraction that multiple UIs implement, but rather have the UIs side-by-side with their own entry points for users and completely independent. Finding common elements and extracting them could be done later.
Users wouldn't access it via `dap.` but rather `dap.uis.<variant>.`

I was thinking one or two UI variants could be built-in, kind of as a demo, but also to improve the out-of-box experience.

PR #57, opened 2020-09-30
Add 'pause' functionality

theHamsta:
I needed to pause a long running job today.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -460,7 +460,11 @@ function Session:event_stopped(stopped)
       if not current_frame then
         return
       end
-      jump_to_frame(current_frame, stopped.preserveFocusHint)
+      local preserve_focus
+      if stopped.reason ~= 'pause' then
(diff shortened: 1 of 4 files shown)


lua/dap.lua
```diff
@@ -460,7 +460,11 @@ function Session:event_stopped(stopped)
       if not current_frame then
         return
       end
-      jump_to_frame(current_frame, stopped.preserveFocusHint)
+      local preserve_focus
+      if stopped.reason ~= 'pause' then
+        preserve_focus = stopped.preserveFocusHint
+      end
+      jump_to_frame(current_frame, preserve_focus)
```
theHamsta, line 467: I always jump to frame on `pause`. My debug adapter sends `preserveFocusHint=true`. But if we only pause a single thread explicitly we probably want to jump to the respective frame.

mfussenegger (approved):

PR #62, opened 2020-10-08
Fix #61: Always ensure a window of the current tabpage is used on stop

### LINKED ISSUE #61: nvim-dap doesn't show buffer on stop when it's opened in a background tab
When you have the buffer where nvim-dap wants to set it's stopped cursor is opened in background tab, you might not notice it. This can be very confusin. It would be better when nvim-dap would re-use a window in the current tab as it would do when the buffer had no active window with it.

`nvim_tabpage_list_wins(0)` instead of `nvim_list_wins` here https://github.com/theHamsta/nvim-dap/blob/0fa0112f0512bbaf76daa3072a422de05a7031a9/lua/dap.lua#L346-L355

I will create a PR for this.

theHamsta:
When nvim-dap uses one of the windows in a background tab the user might not see that the current line cursor was set.

Maybe we could even open the window in a split if no non-special window was found in current tabpage (e.g. repl is the only window). But I don't know whether this would be expected.
```lua
  for _, win in pairs(api.nvim_tabpage_list_wins(0)) do
    local winbuf = api.nvim_win_get_buf(win)
    if api.nvim_buf_get_option(winbuf, 'buftype') == '' then
      local bufchanged, _ = pcall(api.nvim_win_set_buf, win, bufnr)
      if bufchanged then
        api.nvim_win_set_cursor(win, { frame.line, frame.column - 1 })
        with_win(win, api.nvim_command, 'normal zv')
        return
      end
    end
  end
-- after that loop a split?
```

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -352,7 +352,7 @@ local function jump_to_frame(frame, preserve_focus_hint)
   if frame.column == 0 then
     frame.column = 1
   end
-  for _, win in pairs(api.nvim_list_wins()) do
+  for _, win in pairs(api.nvim_tabpage_list_wins(0)) do
     if api.nvim_win_get_buf(win) == bufnr then
       api.nvim_win_set_cursor(win, { frame.line, frame.column - 1 })
       with_win(win, api.nvim_command, 'normal zv')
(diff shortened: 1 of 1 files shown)


mfussenegger:
> Maybe we could even open the window in a split if no non-special window was found in current tabpage (e.g. repl is the only window). But I don't know whether this would be expected.

If another tab has the buffer already open I think I'd expect that tab/window to gain focus.
But if it isn't I think opening a new split window would be a good option.


Do you want to follow up on that in a separate PR? I think this one could already be merged 👍

PR #65, opened 2020-10-29
Allow default setting for excepions.

haringsrob:
Hi,

This is my first lua code ever (not that there is much). But I wanted to be able to set the default exception setting without first starting a session.

Maybe this could be done in a simpler way and just some documentation updates would be required, but this made it work for me.

Happy to hear feedback.

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -171,6 +171,29 @@ Things to note:
 
 
 ==============================================================================
+OTHER CONFIGURATION                                *dap-other-configuration*
+
+If you want to set the exception settings before the debugging session starts
+you can use:
+
+>
+
+    -- Stop on everything (this is the default)
+    require('dap').exception_breakpoints = 'default'
+
+    -- Do not stop on exceptions.
+    require('dap').exception_breakpoints = {}
+
+    -- Ask on session start for settings.
+    require('dap').exception_breakpoints = nil
+
+    -- Stop on specific ones only.
+    require('dap').exception_breakpoints = {'uncaught', '...'}
+
+<
+
+
+==============================================================================
 DEBUGEE CONFIGURATION via launch.json                *dap-launch.json*
 
 nvim-dap supports a small subset of the `launch.json` file that is used to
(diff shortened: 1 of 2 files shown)


doc/dap.txt
```diff
@@ -170,6 +170,29 @@ Things to note:
...
+
+    -- Do not stop on exceptions.
+    require('dap').exception_breakpoints = {}
+
+    -- Ask on session start for settings.
+    require('dap').exception_breakpoints = nil
+
+    -- Stop on specific ones only.
+    require('dap').exception_breakpoints = {'uncaught', '...'}
+
```
theHamsta, line 192: I guess this needs to be a lanugage specific setting. @mfussenegger could we add the debugged language or the debug adapter name to the session table? Then, this setting and custom_events could react language specific.

For this problem it's probably best to set the default_exception settings from the `launch_config`.

mfussenegger: Yes I think we could. Or maybe we could just get the filetype from the buffer.

How about a layered approach. Something like:

```lua
dap.defaults = setmetatable(
  {
    all = {
      exception_breakpoints = `default`;
    },
  },
  {
    __index = function(tbl, key)
      rawset(tbl, key, {}) -- maybe even with a nested metatable here to lookup in `all` ?
      return rawget(tbl, key)
    end
})
```


And you can override it either per language:

```lua
dap.defaults.python.exception_breakpoints = '...'
```

Or global:

```lua
dap.defaults.all.exception_breakpoints = '...'
```

theHamsta: >  Or maybe we could just get the filetype from the buffer.

I guess this will not always work. The user might be in a completely different buffer when he's start a particular debug server. Probably, the best to associate the default break points with a particular config. Then, it's adapter specific and also session specific.

mfussenegger: Created a variant of this in https://github.com/mfussenegger/nvim-dap/pull/67 - what do you think?

mfussenegger:
I merged https://github.com/mfussenegger/nvim-dap/pull/67 - hope that works. 

Thank you for the PR to get this started.

PR #88, opened 2021-01-23
Add variables hover window

mfussenegger:
Continuation of https://github.com/mfussenegger/nvim-dap/pull/44


theHamsta:
I noticed that the new hover window is modifiable. So you can accidentally delete parts of the buffer (not to bad since the window is redrawn on enter)

mfussenegger:
> I noticed that the new hover window is modifiable. So you can accidentally delete parts of the buffer (not to bad since the window is redrawn on enter)

Yes I removed the readonly flag because otherwise I got a warning `W10: Warning: Changing a readonly file` as soon as I opened the hover window.

Will see how I can re-add the readonly behavior without getting the warning.

PR #97, opened 2021-02-02
variables: Enable dynamic evaluation with customizable function

theHamsta:
New default for evaluation expression is visual selection or `cword`

This enable inspecting of member functions without explicit `this`

Screenshot_20210202_203301

Or evaluation from visual selection

image
when selecting `this->m_member`
image

The functions for visual selection come from nvim-treesitter. I wrote the one that gives you the text from the range.

You can replace the function for the current expression to evaluate if you want to.

### DIFF
--- a/lua/dap/ui/variables.lua
+++ b/lua/dap/ui/variables.lua
@@ -155,7 +155,7 @@ local function create_variable_buffer(buf, win, session, root_variables)
   }
   for _, v in pairs(root_variables) do
     -- Is variable expandable?
-    if v.variablesReference > 0 then
+    if v.variablesReference and v.variablesReference > 0 then
       variable_buffers[buf].expanded[v.variablesReference] = true
       if not v.variables then
         session:request(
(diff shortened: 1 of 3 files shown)


lua/dap/ui/variables.lua
```diff
@@ -230,20 +233,32 @@ function M.hover()
...
         break
       end
     end
+
     if variable then
       floating_win, floating_buf = popup()
       create_variable_buffer(floating_buf, floating_win, session, {variable})
     else
-      print('"'..cword..'" not found!')
+      session:evaluate(expression, function(err, resp)
```
mfussenegger, line 250: oh nice :+1:

mfussenegger (approved):
Nice work :+1:

mfussenegger (commented):
One thing that may be a bit strange is that selection marks are not really cleared, so you may get the visual selection behavior even if you want to have the word under the cursor.

Might make sense to check the mode or otherwise add a dedicated `visual_hover` variant?

theHamsta:
I had a solution with `nvim_get_mode` but this somehow does not really work in Lua mode.

theHamsta:
I'll add visual_hover...

PR #99, opened 2021-02-02
Nat sort

theHamsta:
Default sorting in the variable window is by character. This can mess up the order or array indices

Screenshot_20210202_215935

I don't know whether we should switch to a natural sort order by padding all numbers in variable names with zeros???
image

### DIFF
--- a/lua/dap/ui/variables.lua
+++ b/lua/dap/ui/variables.lua
@@ -39,7 +39,13 @@ local function write_variables(buf, variables, line, column, win)
   table.sort(
     sorted_variables,
     function(a, b)
-      return a.name < b.name
+      local num_a = string.match(a.name, '^%[?(%d+)%]?$')
+      local num_b = string.match(b.name, '^%[?(%d+)%]?$')
+      if num_a and num_b then
+        return tonumber(num_a) < tonumber(num_b)
+      else
+        return a.name < b.name
+      end
     end
   )
   local max_textlength = 0


mfussenegger:
Maybe while looping through the variables to build the list to sort it could try to detect if it's is all numbers or alphanumeric and then sort based on that information? 

The padding variant feels a bit like a hack.

theHamsta:
Maybe I need to think more about that. Python pads with zeros automatically. Delve and LLDB have brackets around the numbers.

theHamsta:
I wonder whether we can just pad `^\[\d+\]$`? The simplest solutions i found about natural sorting involved this padding trick since you have to consider consecutive digits as if they were a single token. We get a hint of the parent node how many indexed variables we have to expect but not which of them are indexed.

mfussenegger:
> I wonder whether we can just pad `^\[\d+\]$`? The simplest solutions i found about natural sorting involved this padding trick since you have to consider consecutive digits as if they were a single token. We get a hint of the parent node how many indexed variables we have to expect but not which of them are indexed.
> 
> We could also try to cast strings to numbers in the sorting and allow trimming brackets during that step.


Maybe the sort could try to match `^\[\d+\]$` and then do a `tonumber` on it.

lua/dap/ui/variables.lua
```diff
@@ -32,14 +32,15 @@ local function write_variables(buf, variables, line, column, win)
...
+  local function padzeros(d) return ("%010d"):format(d) end
   local sorted_variables = {}
   for _, v in pairs(variables) do
     table.insert(sorted_variables, v)
   end
   table.sort(
     sorted_variables,
     function(a, b)
-      return a.name < b.name
+      return a.name:gsub("^%[?(%d+)%]?$", padzeros) < b.name:gsub("^%[?(%d+)%]?$", padzeros)
```
mfussenegger, line 43: I had something like this in mind:


```suggestion
      local num_a = string.match(a.name, '^%[?(%d+)%]?$')
      local num_b = string.match(b.name, '^%[?(%d+)%]?$')
      if num_a and num_b then
        return tonumber(num_a) < tonumber(num_b)
      else
        return a.name < b.name
      end
```

It would avoid the arbitrary padding to 10.

PR #100, opened 2021-02-02
variables: add visual_hover for evaluating visual selection

theHamsta:
See https://github.com/mfussenegger/nvim-dap/pull/97#pullrequestreview-581756680


mfussenegger (approved):
> Would evaluate_visual_selection be a better name?

Not sure. I'd keep it as is.

PR #103, opened 2021-02-06
Add a trigger character based autocompletion extension

mfussenegger:
This is quite minimal, all it does is trigger the omnifunc if any of the
trigger characters is typed in insert mode.

The dap omnifunc implementation is async, so this is non-blocking and
should be fast.

This should provide a good experience without dap turning into a full
blown autocompletion plugin.

It is opt-in so users could also use their prefered completion plugin with the omnifunction

https://user-images.githubusercontent.com/38700/107116178-c889cf80-6871-11eb-80d3-02a558ef21d1.mp4

### DIFF
--- a/lua/dap/ext/autocompl.lua
+++ b/lua/dap/ext/autocompl.lua
@@ -1,0 +1,58 @@
+local M = {}
+local api = vim.api
+local dap = require('dap')
+local timer = nil
+
+
+local function destroy_timer()
+  if timer then
+    timer:stop()
+    timer:close()
+    timer = nil
+  end
+end
+
+
+local function trigger_completion()
+  dap.omnifunc(1, "")
+end
+
+
+function M._InsertCharPre()
+  destroy_timer()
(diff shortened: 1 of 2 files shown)


lua/dap/ext/autocompl.lua
```diff
@@ -0,0 +1,57 @@
...
+  destroy_timer()
+  local char = api.nvim_get_vvar('char')
+  if tonumber(vim.fn.pumvisible()) == 1 then
+    return
+  end
+  local session = dap.session()
+  if not session then
+    return
+  end
+  local triggers = session.capabilities.completionTriggerCharacters or {'.'}
```
theHamsta, line 31: Maybe add here a check for `sessions.capabilities` being non-nil. I'm getting error when the session failed to start and therefore has no capabilities.

mfussenegger: Thank you for pointing out :+1: 

Adapted it to handle the nil case.

PR #105, opened 2021-02-06
variables: add scopes hover

theHamsta:
Add a hover window for variable scopes

image

### DIFF
--- a/lua/dap/ui/variables.lua
+++ b/lua/dap/ui/variables.lua
@@ -217,7 +217,7 @@ function M.resolve_expression()
 end
 
 
-function M.hover(resolve_expression_fn)
+local function is_stopped_at_frame()
   local session = require('dap').session()
   if not session then
     print("No active session. Can't show hover window")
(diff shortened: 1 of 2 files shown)


lua/dap/ui/variables.lua
```diff
@@ -217,7 +217,7 @@ function M.resolve_expression()
 end
 
 
-function M.hover(resolve_expression_fn)
+local function do_hover_available_checks()
```
mfussenegger, line 220:
```suggestion
local function check_hover_availability()
```

or

```suggestion
local function is_stopped_at_frame()
```

or

```suggestion
local function is_stopped_with_frame()
```


Not sure what's the best way to name this, but most methods do something, so the `do_` prefix feels a bit redundant and vague.

mfussenegger (commented):
Left a comment with some naming suggestions, otherwise lgtm

mfussenegger (approved):

PR #106, opened 2021-02-06
Variable frames

theHamsta:
This is a follow up of #105 showing the variables of all frames as a tree.

Screenshot_20210207_002610

Some debug adapter can evaluate the variables of different frames while without switching the current frame. Might be useful when browsing for something specific across many frames
Screenshot_20210207_003106

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -574,6 +574,16 @@ hover()                                               *dap.ui.variables.hover()*
 
         Requires an active debug session.
 
+scopes()                                            *dap.ui.variables.scopes()*
+        Opens a hover with all variable scopes of the current frame.
+
+        Requires an active debug session.
+
+frames()                                            *dap.ui.variables.frames()*
+        Opens a hover with the variable of all frames.
+
+        Requires an active debug session.
+
 visual_hover()                                 *dap.ui.variables.visual_hover()*
         Opens a floating window with information on the value of the 
         current visual selection.
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -580,7 +580,7 @@ scopes()                                            *dap.ui.variables.scopes()*
         Requires an active debug session.
 
 frames()                                            *dap.ui.variables.frames()*
-        Opens a hover with the variable of all frames.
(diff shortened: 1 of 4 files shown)


theHamsta:
I noticed that when using debugpy, you might not always get all scopes for all frames. Apparently, debugpy sometimes looses some requests, when makeing them for all frames at once (I observed the same thing with `nvim-dap-virtual-text` which also sometimes looses frames in the `'all frames'` mode)

bew:
Note: didn't read the code.

Is it possible to not have this buffer in a floating window but in a normal window (eg: a vsplit), in case it's very big and want to see the code at the same time?
What's the given liberty on how the window opens?

theHamsta:
@bew there is nothing impeding the use in a normal window. The actual function expects just a buffer and row/col to write the contents into. There was even a split window functionality. But we decided that users expect that the variables in such a watch window would be updated so we deferred this functionality for a later PR since the buffer must subscribe to session events.

But it's totally possible.

mfussenegger:
I wonder if it would be possible to de-couple some of the aspects of the current variable hover feature and then figure out a way how to compose these aspects together in a easy to use API.

E.g to split it into

1. Where to display the information (buffer and window)
2. What to display (Just a table with the data plus some render function?)
3. Actions to take (Expand, but maybe also have an option to have a dialog to pick other actions, like pausing a thread or something like that)
4. Watching for updates


Maybe taking a bit of inspiration from languages like Elm which split things into a view, model and an update effect.

I'm also curious where the [vim.ui](https://github.com/mjlbach/neovim-ui) module is heading.

If we had some way to plug different data and different update effects together then it would be quite easy to also implement a sidebar for the threads or stacktraces. Or to have individual watches for expressions.

theHamsta:
Well, to original draft had this split-up at least between the kind of window and what data to display and the actual render function. I would give up on this PR and just add the update functionality to the variable tree. This would be only an internal rewiring since in the hover you will probably not need automatic updating. Then, it would be a good moment to split up concepts like hover, locals and watch window vs a rendering module that can be replaced.

PR #107, opened 2021-02-07
fix: do not pass vim.empty_dict() to uv.spawn

theHamsta:
The change to use vim.empty_dict() was necessary for jobstart but
not for uv.spawn. This prevented me from starting debugpy.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -955,7 +955,7 @@ function Session:spawn(adapter, opts)
     args = adapter.args;
     stdio = {stdin, stdout, stderr};
     cwd = options.cwd;
-    env = non_empty(options.env) and options.env or vim.empty_dict();
+    env = options.env;
     detached = true;
   }, onexit)
   assert(handle, 'Error running ' .. adapter.command .. ': ' .. pid_or_err)


mfussenegger:
Change itself looks good, though I wonder why I didn't run into this problem. I had also tried with earlier neovim versions to ensure the change in nvim-dap worked before and after the change in neovim.

Which `debugpy` and neovim version are you using?

theHamsta:
I think the main issue with debugpy was that I did not pass any `env`. My neovim is from today
```
❯ nvim --version
NVIM v0.5.0-dev+1071-ge455f0ba2
```
but it has been like that for a while.
```
❯ pip3 list  | rg debugpy
debugpy                      1.2.1
```

mfussenegger (approved):
Can't reproduce the issue. Both `vim.empty_dict()` and `{}` seem to work for the empty case.
But it looks like the `env` option requires a list of "KEY=val" strings, so using `vim.dict_empty()` indeed doesn't make sense.


In case you're curious, I tested it with this:

```
function M.test_env(env)
  local uv = vim.loop
  local stdin = uv.new_pipe(false)
  local stdout = uv.new_pipe(false)
  local stderr = uv.new_pipe(false)
  local handle
  local function onexit()
    stdin:close()
    stdout:close()
    stderr:close()
    if handle then
      handle:close()
    end
  end
  local pid_or_err
  handle, pid_or_err = uv.spawn('/usr/bin/env', {
    args = {};
    stdio = {stdin, stdout, stderr};
    env = env;
    detached = true;
  }, onexit)
  assert(handle, 'Error running /usr/bin/env: ' .. pid_or_err)
  stdout:read_start(function(err, chunk)
    assert(not err, err)
    if chunk then
      print(chunk)
    end
  end)
  stderr:read_start(function(err, chunk)
    assert(not err, err)
    if chunk then
      print(chunk)
    end
  end)
end
```

PR #108, opened 2021-02-07
variables: subscribe_to_session_events to update the variable buffer

theHamsta:
@bew this is a draft to update the variable buffer whenever necessary. So it would be possible to have a variable window in a split and it will update the values of the variables. you can try this by opening the hover in a split with `<c-w>v`

Very experimental at the moment.

TODO
- [ ] also update variable windows that were created from an evaluate request
- [ ] update regular variables on evaluate requests that could have side effects?
- [ ] check whether it's possible to keep the expanded/not expanded state when variables change
- [ ] dedup updates on `variables` request

### DIFF
--- a/lua/dap/ui/variables.lua
+++ b/lua/dap/ui/variables.lua
@@ -1,6 +1,7 @@
 local api = vim.api
 local utils = require("dap.utils")
 local non_empty = utils.non_empty
+local dap = require'dap'
 
 local M = {}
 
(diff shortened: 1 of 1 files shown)


lua/dap/ui/variables.lua
```diff
@@ -277,7 +346,7 @@ function M.hover(resolve_expression_fn)
             floating_win, floating_buf = popup()
             resp.value = resp.result
             resp.name = expression
-            create_variable_buffer(floating_buf, floating_win, session, {resp})
+            create_variable_buffer(floating_buf, floating_win, session, function() return {resp} end)
```
theHamsta, line 349: this is not dynamic at the moment

theHamsta:
Also works across different sessions.

theHamsta:
I guess https://github.com/rcarriga/nvim-dap-ui has solved this problem, thanks!

PR #114, opened 2021-03-03
refactor: Expose modules in multiple files and handle global state better

tjdevries:
Hey, I'm not quite done with this, but figured I should probably ask before I do too much stuff....

Would you accept a PR that does this? It should fix a couple things:

1. It will make it so you can reload the files while executing and keep global state
2. Should reduce global dependency on variables and parameterize more functions
3. Should allow more easy re-use by plugins since they'll be able to access a few more of the pieces
4. Hopefully easier to read & follow since things will be more isolated and single purpose.

If you don't like it, that's fine too :)

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -3,34 +3,23 @@ dap = {} -- luacheck: ignore 111 - to support v:lua.dap... uses
 
 local api = vim.api
 
-local log = require('dap.log').create_logger('dap.log')
-local repl = require('dap.repl')
 local dap_signs = require('dap.signs')
+local log = require('dap.log').create_logger('dap.log')
+local reloadable = require('dap.reloadable')
+local repl = require('dap.repl')
 local ui = require('dap.ui')
 local utils = require('dap.utils')
-
-local non_empty = utils.non_empty
-
-local M = {}
-local last_run = nil
-
-local reloadable = require('dap.reloadable')
-
 local Session = require('dap.session')
 
+local non_empty = utils.non_empty
 local ns_breakpoints = require('dap.constants').ns_breakpoints
 
-local function set_current_session(val)
-  reloadable.set_value('CurrentSession', val)
-
-  return val
-end
-
-local function get_current_session()
-  return reloadable.get_value('CurrentSession')
-end
+local set_current_session, get_current_session = reloadable.create_value('CurrentSession')
+local set_last_run, get_last_run = reloadable.create_value('LastRun')
 
 local bp_info = reloadable.table('BpInfo')
+
+local M = {}
 
 M.repl = repl
 M.custom_event_handlers = setmetatable({}, {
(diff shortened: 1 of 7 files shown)


mfussenegger (commented):
In general it sounds good to me to split things up a bit more. `dap.lua` has grown quite a bit.


Just writing down my first reactions/thoughts, don't take this as judgement:


- rpc: Makes sense to me
- session: Makes sense to me, although I'm a bit worried about exposing all the functions as it makes it harder to avoid breaking something accidentally. But I can see how it is appealing and useful. So I think it's good to go
- contants: Kinda makes sense, but also feels a bit overkill to extract this. Is to avoid circular imports? Not really opposed to it since it has a clear scope.
- signs: Not sure if it's about signs or if it should be about breakpoints? Both seems valid.
- reloadable: Not sure I fully grok what it is for. Do I understand it right that you want to put all the state into it, so that you can reload the other modules without discarding the state?

tjdevries:
> In general it sounds good to me to split things up a bit more. `dap.lua` has grown quite a bit.
> 
> Just writing down my first reactions/thoughts, don't take this as judgement:
> 
>     * rpc: Makes sense to me
> 
>     * session: Makes sense to me, although I'm a bit worried about exposing all the functions as it makes it harder to avoid breaking something accidentally. But I can see how it is appealing and useful. So I think it's good to go

I think you can say "session object's methods considered private :)"

>     * contants: Kinda makes sense, but also feels a bit overkill to extract this. Is to avoid circular imports? Not really opposed to it since it has a clear scope.

Yeah, literally just for circular imports things. Otherwise I wouldn't have made it.

>     * signs: Not sure if it's about signs or if it should be about breakpoints? Both seems valid.

Yeah, no strong preference there either, was just trying to move those functions somewhere. Probably makes more sense to name it breakpoints and move all breakpoints things there.

>     * reloadable: Not sure I fully grok what it is for. Do I understand it right that you want to put all the state into it, so that you can reload the other modules without discarding the state?

This is so you can do:

```lua
package.loaded['dap'] = nil
require('dap')
```

to re-execute the file WITHOUT losing state. So you still have your session, breakpoints, etc. saved. This makes it many many times faster to iterate on UI / UX without restarting nvim or even restarting the debugging session. I'm sure I haven't hit all the spots yet, but this will greatly improve playing around with user experience stuff.

mfussenegger:
@tjdevries do you mind if I continue with this?

tjdevries:
@mfussenegger go for it -- I can always submit PRs again after you finish if I see anything :) I don't have a bunch of time right now. I pushed some random changes I had just made as well.

mfussenegger:
I split up some of the components. There's no `state` or `reloadable` module yet because reloading only parts can be problematic if there are references held that point to the old code.

I'm currently using something like this:

```
function M.reload_dap()
  require('dap.repl').close()
  U.reload('dap', true)
  U.reload('me.dap.conf')
  require('me.dap.conf').setup()
  U.reload('jdtls.dap').setup_dap()
end
```

To develop on the plugin (and tests(!)) which seems good enough for the time being.
So I'm going to close the PR for now. Thanks again for pushing this :)

PR #116, opened 2021-03-03
Workaround for empty setBreakpoints response

mfussenegger:
Looks like `delve` can send a response that is `nil`:

```
.config/nvim/pack/plugins/start/nvim-dap/lua/dap.lua:760: attempt to index local 'resp' (a nil value)
```

This is with the following adapter config:

```lua
  dap.adapters.go = {
    type = 'server',
    host = '127.0.0.1',
    port = 46545
  }
```

https://user-images.githubusercontent.com/38700/109873407-fcd48e00-7c6d-11eb-8eef-108b8a3d56c2.mp4

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -756,13 +756,18 @@ function Session:set_breakpoints(bufexpr, on_done)
     self:request('setBreakpoints', payload, function(err1, resp)
         if err1 then
           print("Error setting breakpoints: " .. err1.message)
-        else
+        elseif resp then
           for _, bp in pairs(resp.breakpoints) do
             if not bp.verified then
               log.info('Server rejected breakpoint', bp)
               remove_breakpoint_signs(bufnr, bp.line)
             end
           end
+        else
+          log.warn(
+            'setBreakpoints response contains neither an error nor a body. ' ..
+            'This is likely a bug in the debug adapter.'
+          )
         end
         num_bufs = num_bufs - 1
         if num_bufs == 0 and on_done then


theHamsta:
Yeah, I changed this also sometimes but was not sure whether I was doing something wrong. Probably good to check from this case.

mfussenegger:
I'm actually not sure what is really going on.

From the log it looks like the response is sent twice:

```
[ DEBUG ] 2021-03-03T22:47:38Z+0100 ] ...rdi/.config/nvim/pack/plugins/start/nvim-dap/lua/dap.lua:837 ]	{
  body = {
    breakpoints = { {
        line = 7,
        source = {},
        verified = true
      } }
  },
  command = "setBreakpoints",
  request_seq = 1,
  seq = 0,
  success = true,
  type = "response"
}
[ WARN ] 2021-03-03T22:47:38Z+0100 ] ...rdi/.config/nvim/pack/plugins/start/nvim-dap/lua/dap.lua:842 ]	"No callback for "	{
  body = {
    breakpoints = { {
        line = 7,
        source = {},
        verified = true
      } }
  },
  command = "setBreakpoints",
  request_seq = 1,
  seq = 0,
  success = true,
  type = "response"
}
```

The first clears the callback, so the second time there is the `No callback for ` warning.
What I don't get is that although there is a `body` in the log, it somewhere goes missing.

An assertion like this:

https://github.com/mfussenegger/nvim-dap/blob/7dc11c3c8a59950e2e04bbd319dfa693bb41ba5d/lua/dap.lua#L847-L848

Triggers an error.

mfussenegger:
Looks like the root cause of the issue is that `seq` is not being increased in delve, and nvim-dap currently relies on that for the number increments. So the `request_seq` always stays the same.

See https://github.com/go-delve/delve/issues/2245 for a bug report

mfussenegger:
https://github.com/mfussenegger/nvim-dap/pull/117 is the proper fix

Issue #119, opened 2021-03-10
The backspace key works in a strange way on repl buffer.


pittcat:
https://user-images.githubusercontent.com/22888402/110602154-a8c32f80-81c0-11eb-8a85-8f209bbd3c08.mp4

pittcat:
Enter key and <C-w> do not work as well as they do in terminal.

theHamsta:
I think most of this is due to vim's weird/borken implementation of `prompt_buffer`s that aren't really user friendly. The behavior of backspace might be due to setting `unmodified` `on_lines`??? I personally never could these prompt buffer make working with the look and feel of a normal terminal buffer.

pittcat:
When you enter a variable and then reuse it by pressing `<Up>` key, the `<C-w>` won't take effect on deleting characters. The problem with the enter key is that even if you don't enter any variables, it will still act as if you entered the last variable.

theHamsta:
> The problem with the enter key is that even if you don't enter any variables, it will still act as if you entered the last variable.

It will repeat the last command that you gave it. Many command line debugger have this behavior for stepping command. However, pdb does not repeat an eval request just commands. GDB repeats every command.

```
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./a.out...
(gdb) break main
Breakpoint 1 at 0x11c9: file main.cpp, line 17.
(gdb) 
Note: breakpoint 1 also set at pc 0x11c9.
Breakpoint 2 at 0x11c9: file main.cpp, line 17.
(gdb) w
Missing arguments.
(gdb) f
No stack.
(gdb) l
warning: Source file is more recent than executable.
3        * Copyright (C) 2021 Stephan Seitz <stephan.seitz@fau.de>
4        *
5        * Distributed under terms of the GPLv3 license.
6        */
7
8       #include <cstdlib>
9       #include <iostream>
10
11      struct Foo {
12        int x = 2;
(gdb) n
The program is not being run.
(gdb) r
Starting program: /home/stephan/a.out 

Breakpoint 1, main (argc=21845, argv=0x5555555552b0 <__libc_csu_init>) at main.cpp:17
17        std::cout << a << std::endl;
(gdb) n
18        return a + b + 2;
(gdb) argc
Undefined command: "argc".  Try "help".
(gdb) p argc
$1 = 1
(gdb) 
$2 = 1
(gdb) 
```

mfussenegger:
[...]

PR #120, opened 2021-03-10
Don't reset modified flag on REPL input

mfussenegger:
Might solve https://github.com/mfussenegger/nvim-dap/issues/119

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -200,7 +200,7 @@ function M.close()
   end
 
   if buf then
-    vim.cmd('bd' .. buf)
+    api.nvim_buf_delete(buf, {force = true})
     buf = nil
   end
 
@@ -242,12 +242,6 @@ function M.open(winopts, wincmd)
     vim.fn.prompt_setprompt(buf, 'dap> ')
     vim.fn.prompt_setcallback(buf, execute)
     api.nvim_buf_attach(buf, false, {
-      on_lines = function(b)
-        api.nvim_buf_set_option(b, 'modified', false)
-      end;
-      on_changedtick = function(b)
-        api.nvim_buf_set_option(b, 'modified', false)
-      end;
       on_detach = function()
         buf = nil
         return true


theHamsta:
I personally find the `did not save "[dap-repl]"` messages much more annoying than not being able to use `<bs>`. But that's just a personal opinion. We found no other way to tell Vim that this is not a normal file that needs to be saved the last time, didn't we?

mfussenegger:
> We found no other way to tell Vim that this is not a normal file that needs to be saved the last time, didn't we

I don't think so,  I'll have another look and otherwise file something in neovim core

mfussenegger:
https://github.com/vim/vim/commit/4551c0a9fcdbdef52836d4852686d54b5e47fdaf makes it sound as the behavior was somewhat intentional?

Maybe an option would be do the `api.nvim_buf_set_option(buf, 'modified', false)` call less often? E.g. on `WinLeave` or something like that.

theHamsta:
That's why we have Neovim. To not be dependent on someone who wants some weird behavior and ends up hard-coding it into C. Pretty strange though that this is not configurable. Well it makes sense if you don't use `set bufhidden=hide`. But most people do.

With the events we did not hit every change of the buffer.

theHamsta:
In vlime (a Lisp plugin with REPL), we also tried many combinations of making the buffer non-editable and make it automatically non-modified. Main-problem was that Lisp can only print parts of a line but you're not supposed to modify it. Just text you wrote. But in the end we found no satisfying solution and you still cannot enter text in the REPL. I have the feeling that this can probably only be tackled in vim itself.

theHamsta:
Another bug is that you get only text output in the REPL when you have been at least once in input mode. Otherwise, no output will appear in the prompt buffer.

mfussenegger:
https://github.com/mfussenegger/nvim-dap/pull/123 is probably the better option to workaround the backspace issue. Closing this here.

PR #121, opened 2021-03-11
feat: output and exit handlers

rcarriga:
Hey I'm the maintainer of [vim-ultest](https://github.com/rcarriga/vim-ultest), a UI for running tests using vim-test under the hood. One feature I would love to have is integration with nvim-dap, as I often want to debug while testing. I know nvim-dap-python tries to do test running for python but ideally this integration would work for any language supported by vim-test. To do this, vim-ultest needs to be able to capture the output and exit code of the test. The rest can be handled externally to nvim-dap.

I've done up this simple PR to accomplish this but I'd be completely open to minor changes or totally rethinking this. If this is something that you'd rather handle yourself no worries, I just wanted to suggest something that I have working. 

Here's an example of it in action 

dap

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -586,7 +586,10 @@ function Session:event_terminated()
 end
 
 
-function Session.event_exited()
+function Session:event_exited(args)
+  if self.handlers.exit then
+    self.handlers.exit(args.exitCode)
+  end
 end
 
 function Session.event_module()
@@ -601,8 +604,13 @@ end
 function Session.event_continued()
 end
 
-function Session.event_output(_, body)
-  repl.append(body.output, '$')
+function Session:event_output(body)
+  if self.output_to_repl then
+    repl.append(body.output, '$')
+  end
+  if self.handlers.output and vim.tbl_contains({"stdout", "stderr"}, body.category) then
+    self.handlers.output(body.output)
+  end
 end
 
 function Session:_request_scopes(current_frame)
@@ -878,8 +886,11 @@ end
 local function session_defaults(opts)
   local handlers = {}
   handlers.after = opts.after
+  handlers.exit = opts.exit
+  handlers.output = opts.output
   return {
     handlers = handlers;
+    output_to_repl = opts.output_to_repl == nil or opts.output_to_repl;
     message_callbacks = {};
     message_requests = {};
     initialized = false;


theHamsta:
You know that you can subscribe to all events and request responses of the DAP protocol? 

https://github.com/theHamsta/nvim-dap-virtual-text/blob/master/lua/nvim-dap-virtual-text.lua

rcarriga:
I did not know about that, thanks! 

One issue is that the handlers I need to inject are dynamic i.e. contain some state, such as a test id. This would be possible with the that solution by maintaining some global state but would add a lot more complexity, unless I'm missing something!

theHamsta:
@rcarriga you can set the custom_handlers dynamically. You can set them and delete whenever you want and how many you want.

Each event and response has a dictionary with subscribers with their respective id. They are called right after nvim-dap has handled the event. When you delete yourself from the dictionary your function will not be called. 

```

rcarriga:
Oh I see... I imagined the mutating the table might have some side effects but if that's safe then yeah that could definitely work thank you!

Another issue that was solved by this PR was preventing output to the repl if not wanted. This is useful for instance here because the output of pytest sends colour codes. Is that something that is already possible or would it be something worth submitting separately?

theHamsta:
It's not really possible right now. Of course you could overwrite session.event_output but this would be a bit hacky.

theHamsta:
Maybe it will be possible soon to print those terminal codes correctly: https://github.com/neovim/neovim/pull/13567

mfussenegger:
> Hey I'm the maintainer of vim-ultest, a UI for running tests using vim-test under the hood. One feature I would love to have is integration with nvim-dap, as I often want to debug while testing. I know nvim-dap-python tries to do test running for python but ideally this integration would work for any language supported by vim-test

Sounds good :+1: 


> Another issue that was solved by this PR was preventing output to the repl if not wanted. This is useful for instance here because the output of pytest sends colour codes. Is that something that is already possible or would it be something worth submitting separately?


https://github.com/mfussenegger/nvim-dap/pull/114 would expose the `Session` and then the handler could be overridden.

You could also use the `console = 'integratedTerminal'` option which some debug adapters support, then you'd get the output already in a terminal with color codes support. Although I guess you'd want more control over how the terminal is displayed?


> Maybe it will be possible soon to print those terminal codes correctly: neovim/neovim#13567

I've already had that branch checked out today to try the feature :)

rcarriga:
> You could also use the console = 'integratedTerminal' option which some debug adapters support, then you'd get the output already in a terminal with color codes support. Although I guess you'd want more control over how the terminal is displayed?

Yes I already have a way to display the output to the user so it'd be nice to be able to just disable in the REPL. Also ideally I'd like to avoid touching the adapter config as much as possible to remain agnostic of it. As it's just an aesthetic thing for now, I'm happy waiting for the session to be exposed, that's great it's in the pipeline.

> Maybe it will be possible soon to print those terminal codes correctly: neovim/neovim#13567

That's awesome I didn't know that was happening, that'd be great for my plugin as right now I'm piping everything into `less` which is just a pain to rely on.

Thanks for the help on this, I'll close

mfussenegger:
One more thing that may be worth pointing out:

`dap.run` takes as second argument a table that can have a `before` and `after` function. You could use these to setup and remove the custom event handler. This should then also play well with the `dap.run_last()` method.

See https://github.com/mfussenegger/nvim-jdtls/blob/8ff60d5e91fe2a4c1dedc6685ef7722e8e7bce78/lua/jdtls/dap.lua#L161-L180 for an example where it's used

mfussenegger:
One more thing. Once you've merged your integration, would you create a PR against nvim-dap-python to mention vim-ultest in the readme as alternative? And you could also add it to https://github.com/mfussenegger/nvim-dap/wiki/Extensions  :)

rcarriga:
> One more thing that may be worth pointing out:
> 
> `dap.run` takes as second argument a table that can have a `before` and `after` function. You could use these to setup and remove the custom event handler. This should then also play well with the `dap.run_last()` method.
> 
> See https://github.com/mfussenegger/nvim-jdtls/blob/8ff60d5e91fe2a4c1dedc6685ef7722e8e7bce78/lua/jdtls/dap.lua#L161-L180 for an example where it's used

Yeah that's exactly what I need :+1: 

> One more thing. Once you've merged your integration, would you create a PR against nvim-dap-python to mention vim-ultest in the readme as alternative? And you could also add it to https://github.com/mfussenegger/nvim-dap/wiki/Extensions :)

Absolutely will do!

PR #123, opened 2021-03-12
Ensure backspace works in REPL

mfussenegger:
Workaround for https://github.com/mfussenegger/nvim-dap/issues/119


mfussenegger:
This workaround also doesn't fully work as it might switch to normal mode. Closing this.

Not sure if there's a good workaround at all - probably needs to be fixed upstream to work without hassles

PR #124, opened 2021-03-14
fix: call custom events first

rcarriga:
Following on from the discussion on #121, I've attempted to do something like
```lua
      before = function(config)
        dap.custom_event_handlers.event_output[handler_id] = output_handler
        dap.custom_event_handlers.event_terminated[handler_id] = terminated_handler
        dap.custom_event_handlers.event_exited[handler_id] = exit_handler
        return config
      end,
      after = function()
        dap.custom_event_handlers.event_exited[handler_id] = nil
        dap.custom_event_handlers.event_terminated[handler_id] = nil
        dap.custom_event_handlers.event_output[handler_id] = nil
      end
```
However, nvim-dap also handles the terminated event and calls the `after` function which means the terminated handler is removed before being used. So I just changed it to call custom handlers first, though I'm not sure if this will break something in other plugins that rely on dap having done some handling first... Any concerns?

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -859,10 +859,10 @@ function Session:handle_body(body)
     local callback = self['event_' .. decoded.event]
     if callback then
       vim.schedule(function()
-        callback(self, decoded.body)
         for _, c in pairs(M.custom_event_handlers['event_' .. decoded.event]) do
           c(self, decoded.body)
         end
+        callback(self, decoded.body)
       end)
     else
       log.warn('No event handler for ', decoded)


rcarriga:
I could of course, treat the `after` function as the terminated event, but that means the code must maintain the knowledge of how nvim-dap uses the after function. Feels like that this solution is most consistent

theHamsta:
The rationale of calling the custom event after nvim-dap handles is that the session is brought to the right state in which it should be after the corresponding event/response. E.g. it should have all variables after a variables request and the frames after a frames request or in the stopped state after a stopped event. This is important when you want to use functions or members of session. E.g. my virtual text plugin can access the sessions variables and must not do own requests or track the variables (a variables response only gives you new variables, not all variables that have been evaluated)

I don't really get what you want to say with this after/before function? Is this a concept of you test runner? Each handler can also remove it self when it's running and doesn't want to be called again. In https://github.com/mfussenegger/nvim-dap/pull/108/files I'm unsubscribing myself on buffer detach and I'll have a per-buffer subscription.

mfussenegger:
> I'm not sure if this will break something in other plugins that rely on dap having done some handling first

I'm pretty sure it would if a plugin depends on the current behavior.


Although the `terminated_handler` could de-register itself as theHamsta pointed out, I think it would make sense to give users the choice to register listeners that trigger either before or after the internal handlers.


We could have a `dap.listeners.before` and `dap.listeners.after`, both with the same structure as the current `custom_event_handlers` and `custom_response_handlers`. With response/event merged together, because the keys are unique anyways.

This would also make it explicit when something is called.

We'd have to add a deprecation warning for the existing extension points, and add some fallback to the new `dap.listeners.before` and `dap.listeners.after` so that things continue working (+ warning) until we remove it.

theHamsta:
Could you explain my what advantage it could have to be called before nvim-dap?

As far as I understood @rcarriga wants to register two functions where he could just one and rely that one of these functions get called before the other one (the action must run before the unsubscribe). To put the handler before to just have delayed effect for the unsubscribe action seems difficult for me to document and a very indirect way to achieve the actual goal.

rcarriga:
> I don't really get what you want to say with this after/before function? Is this a concept of you test runner? Each handler can also remove it self when it's running and doesn't want to be called again. In https://github.com/mfussenegger/nvim-dap/pull/108/files I'm unsubscribing myself on buffer detach and I'll have a per-buffer subscription.

The problem with this is that it means all events must be triggered to unregister themselves. This is a problem for instance with vscode-go which doesn't trigger the `exit` event. Of course the differences could be worked around but then extensions need to be aware of the inconsistency.

Although I do like giving users more control with the `before/after` handlers, I can see how that would be a pain with migration and documentation. What about slightly changing the behaviour of the `callback` so that it can return a function to be called after the custom handlers? nvim-dap can then make sure everything is in place for custom handlers while also remaining consistent with the terminated event...

In code this would be something like:

```lua
        after_handlers = callback(self, decoded.body)
        for _, c in pairs(M.custom_event_handlers['event_' .. decoded.event]) do
          c(self, decoded.body)
        end
        if after_handlers then
          after_handlers()
        end

```

```lua
function Session:event_terminated()
  self:close()
  if self.handlers.after then
    return function()
      self.handlers.after()
      session = nil
    end
  else
    session = nil
  end
end

```

I'm probably missing something there with deleting the session but just an initial thought

mfussenegger:
> Could you explain my what advantage it could have to be called before nvim-dap?

I think there are use cases.

For example, you could want to print some information about the session before it ends. E.g. how many threads the application used or something like that.

Right now `event_terminated` couldn't access that information anymore, because it's already reset by the time the custom handler is called.

I could also imagine that it would allow to apply formatting rules to the variables before they're displayed by other components. Although this is also a bit of a downside, because it opens the door for plugins/users to do nasty stuff. 




> What about slightly changing the behaviour of the callback so that it can return a function to be called after the custom handlers?

I'm not sure. It feels a bit like a workaround and special casing for the `event_terminated` / `after` case and it has less utility, at least until the Session handlers can be overridden directly. But once we do that, people who override them would have one more thing to look out for to not accidentally break this `after_handlers` behavior.

I'd rather update the documentation and put in some deprecation logic. The ecosystem that uses the custom handlers isn't too big yet. It's better to do it now than later.

rcarriga:
Sounds good to me!

mfussenegger:
@rcarriga I've created https://github.com/mfussenegger/nvim-dap/pull/126

PR #125, opened 2021-03-14
fix: dapl.repl.run_last to dap.run_last in example mappings

fdschmidt93:
This little line caused me some unfortunate headaches in setting up my mappings :sweat_smile:


mfussenegger:
Oh, thank you for the fix :+1:

PR #126, opened 2021-03-15
Add listeners.{before,after} as replacement for custom_ handlers

mfussenegger:
As discussed in https://github.com/mfussenegger/nvim-dap/pull/124 this
gives users the possible to choose whether the listeners should be
called before or after the internal listeners.

I also used the opportunity to add another change: The response
listeners are now also called in error cases. So the number of arguments
changed.

`custom_event_handlers` and `custom_response_handlers` are both still
there and used. (Without the signature change for request responses)
If used a warning will be printed out.

They'll be removed at some point in the future.

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -622,20 +622,23 @@ debug adapter might send to nvim-dap.
 
 There are two tables for that:
 
-- `custom_event_handlers`
-- `custom_response_handlers`
+- `dap.listeners.before`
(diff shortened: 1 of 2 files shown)


rcarriga:
This is great, I love the concise naming! Thank you!

theHamsta:
I like this idea to have multiple listeners more than the one nvim lsp where you have to make your own handler that multiplexes between different plugins.

PR #130, opened 2021-03-22
feat: Add support for stepInTargets

theHamsta:
Can be tested with: https://github.com/microsoft/debugpy/pull/563 (merged)

It will ask to step into which target if there are more stepInTargets in one line (makes more sense for list comprehensions than having multiple statements in one line). I added an option to not ask for step targets which would correspond to the old behavior.

image

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -1097,12 +1097,18 @@ function Session:_step(step, granularity)
   self.stopped_thread_id = nil
   vim.fn.sign_unplace(ns_pos)
 
+  if step == 'continue' or step == 'reverseContinue' then
+    granularity = nil
+  else
+    granularity = granularity or M.defaults[self.config.type].stepping_granularity
+  end
+
   local function send_request(target_id)
     self:request(step,
     {
       threadId = thread_id;
       targetId = target_id;
-      granularity = granularity or M.defaults[self.config.type].stepping_granularity
+      granularity = granularity
     }, function(err0, _)
       if err0 then
         print('Error on '.. step .. ': ' .. err0.message)
@@ -1121,7 +1127,7 @@ function Session:_step(step, granularity)
         if #response.targets > 1 then
           ui.pick_one(
             response.targets,
-            "Step into what?",
+            "Step into which function?",
             function(target) return target.label end,
             function(target)
               if not target or not target.id then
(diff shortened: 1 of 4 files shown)


lua/dap.lua
```diff
@@ -1186,11 +1187,43 @@ function Session:_step(step)
...
+  end
+
+  if self.capabilities.supportsStepInTargetsRequest
+    and M.defaults[self.config.type].ask_step_in_targets
+    and self.current_frame then
+    self:request("stepInTargets", { frameId = self.current_frame.id }, function(err1, response)
+      if err1 then
+        print('Error on '.. step .. ': ' .. err1.message .. " (while requesting stepInTargets)")
+      else
+        if #response.targets > 1 then
```
theHamsta, line 1206: should the UI change to just select the one entry if there is only one

lua/dap.lua
```diff
@@ -57,6 +57,9 @@ M.defaults = setmetatable(
   {
     fallback = {
       exception_breakpoints = 'default';
+      ask_step_in_targets = true;
+      -- type SteppingGranularity = 'statement' | 'line' | 'instruction'
+      stepping_granularity = 'statement';
```
theHamsta, line 62: Also added SteppingGranularity. This does not require a capability but most adapters will ignore this information.

theHamsta: actually all step request support granularity

theHamsta: except "continue"/"reverseContinue"

lua/dap.lua
mfussenegger: I think it makes sense to just step into without prompt if there is only 1 target. So would keep it as it is.

mfussenegger (commented):
I'll test this a bit but overall looks good. Thank you for this.

Too bad it's not supported yet by the java debug adapter :)

lua/dap.lua
theHamsta: I meant whether we should generalize this to all `ui.pick` decisions

mfussenegger: Oh. I'm not sure. `select_config_and_run` also behaves like that - so I guess it would make sense.

Would you change the default `ui.pick_one` implementation? It's intended to be overriden with fancier solutions, so we have to keep that in mind.
Might make sense to make it more explicit and have a separate method around `ui.pick_one`

mfussenegger:
@theHamsta I appended some changes to this in the last commit. Please let me know what you think.

theHamsta:
Looks good to me! 👍

PR #132, opened 2021-03-24
fix: call read_start once connected

jbyuki:
I have been playing around with nvim-dap and an lua adapter for it. TCP communication got me stuck for a good moment but I think this fixes it. This calls `read_start` once in the `connect` callback. Similar to [here](https://github.com/luvit/luv/blob/master/examples/uvbook/tcp-echo-client.lua).


mfussenegger:
Thank you for the PR :+1:

PR #134, opened 2021-03-31
Add support for data breakpoints

theHamsta:
First draft for data breakpoints.

https://user-images.githubusercontent.com/7189118/113154094-f5d27880-9237-11eb-90b1-cebf23dd701e.mp4

They can persist over sessions. Java debug adapter only supports data breakpoints for value contained in some kind of container (a class or object) not free standing stack variables like a loop counter `i`.


lua/dap.lua
```diff
@@ -50,6 +51,22 @@ M.listeners = {
 }
 
 
+function M.listeners.after.event_initialized.databreakpoints()
+  if session and session.capabilities and session.capabilities.supportsDataBreakpoints then
```
theHamsta, line 55: could move these also in the respective main functions for the respective events

lua/dap/ui/variables.lua
```diff
@@ -99,6 +107,34 @@ local function update_variable_buffer(buf, win)
...
+  local buf = api.nvim_get_current_buf()
+  local state = variable_buffers[buf]
+  local pos = api.nvim_win_get_cursor(0)
+  if not state then return end
+
+  local variable = state.line_to_variable[pos[1] - 1]
+
+  if variable then
+    local parent
+    for _, p in pairs(state.line_to_variable) do
```
theHamsta, line 120: this will only work when the parent is also display in this variable buffer. So if the breakpoint is set on any of the non-root elements. A depth search on `current_frame.variables` might be more precise.

lua/dap.lua
```diff
@@ -1357,6 +1464,30 @@ function M.set_exception_breakpoints(filters, exceptionOptions)
...
 
 
+function M.clear_data_breakpoints()
+  data_bp_info = {}
+  if not session then return end
+  session:_set_data_breakpoints()
+end
+
+
+function M.toggle_data_breakpoint(resolve_expression_fn, condition, hit_condition)
```
theHamsta, line 1474: The DAP protocol allows data breakpoints to be an arbitrary expression. But I don't use an adapter that support this.

lua/dap.lua
mfussenegger: I'd tend to move them into the respective events, given that part of the functionality already requires changes in the built-in event handlers and breakpoints so far are built-in anyways

PR #139, opened 2021-04-08
Fix #138: Make stepInTargets an explicit command

### LINKED ISSUE #138: Make StepInTarget a Separate Function
debugpy has changed its behavior to no longer just provide stepInTargets on the same line but within the whole stack frame. Maybe we should ask for stepInTargets only when the user has explicitly asked for (parameter for `step_into` it or offer it as a separate function. 

image

theHamsta:
We could also keep the current behavior and but only accept target in the same line or the same statement. I just had the feeling that when you can step into any arbitrary function in the stack frame it would be different enough that the user would like to have both normal stepIn and stepIn with targets.

I don't know whether this is the best strategy. The protocol is very flexible on when its about what adapter should return and how debugger should use the response best in its UI.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -40,7 +40,6 @@ M.defaults = setmetatable(
   {
     fallback = {
       exception_breakpoints = 'default';
-      ask_step_in_targets = true;
       -- type SteppingGranularity = 'statement' | 'line' | 'instruction'
       stepping_granularity = 'statement';
     },
@@ -1026,16 +1025,18 @@ function Session:_step(step, params)
   end)
 end
 
-function M.step_over()
+function M.step_over(opts)
   if not session then return end
-  session:_step('next')
+  session:_step('next', opts)
 end
 
-function M.step_into()
+function M.step_into(opts)
   if not session then return end
-  local ask_targets = M.defaults[session.config.type].ask_step_in_targets
-  if not (ask_targets and session.capabilities.supportsStepInTargetsRequest) then
-    session:_step('stepIn')
+  opts = opts or {}
+  local askForTargets = opts.askForTargets
+  opts.askForTargets = nil
+  if not (askForTargets and session.capabilities.supportsStepInTargetsRequest) then
+    session:_step('stepIn', opts)
     return
   end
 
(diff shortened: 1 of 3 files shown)


lua/dap.lua
```diff
@@ -1104,10 +1103,9 @@ function M.step_over()
   session:_step('next')
 end
 
-function M.step_into()
+function M.step_into(ask_for_targets)
```
mfussenegger, line 1106: I wonder if we should make this an `opts` dictionary already to make it easier to extend it without changing the signature or forcing users to pass unused arguments.

Could imagine that we might want to add a function to pre-filter the targets (e.g. to have the "only targets that match the current line" behavior that you mention in the PR description).

But also fine for me to merge as is - let me know what you think.

theHamsta: yes, options is better. It could include also the granularity and more stuff if needed later.

lua/dap.lua
```diff
@@ -1060,16 +1059,17 @@ function Session:_step(step, params)
...
+  session:_step('next', options)
 end
 
-function M.step_into()
+function M.step_into(options)
   if not session then return end
-  local ask_targets = M.defaults[session.config.type].ask_step_in_targets
-  if not (ask_targets and session.capabilities.supportsStepInTargetsRequest) then
-    session:_step('stepIn')
+  local askForTargets = options.askForTargets
```
mfussenegger, line 1069:
```suggestion
  local askForTargets = options and options.askForTargets
```

Otherwise this causes an error if options is nil

lua/dap.lua
```diff
@@ -1087,31 +1087,32 @@ function M.step_into()
         if not target or not target.id then
           print('No target selected. No stepping.')
         else
-          session:_step('stepIn', { target_id = target.id })
+          options.target_id = target.id
```
mfussenegger, line 1090: Same here, might cause an error if `options` is `nil`. Could also add a `options = options or {}` at the start of the method to avoid that.


Btw. should we rename `options` to `opts` ? For consistency as that's already used in other places. (I went with `opts` instead of the more verbose variant because it's also commonly used in the neovim code base)

PR #149, opened 2021-04-12
Prompt user with choices on continue if not stopped at thread

mfussenegger:
I'm not entirely sure if this is a good idea.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -397,8 +397,36 @@ end
 function M.continue()
   if not session then
     select_config_and_run()
-  else
+  elseif session.stopped_thread_id then
     session:_step('continue')
+  else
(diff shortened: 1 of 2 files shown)


lua/dap.lua
```diff
@@ -393,8 +393,32 @@ end
...
+      },
+      {
+        label = "Restart session",
+        action = M.restart,
+      },
+      {
+        label = "Do nothing",
+        action = function() end,
+      },
+    }
```
theHamsta, line 416: maybe even add alternative `pause`? In vscode continue and pause are the same button that toggles depending on current session state

mfussenegger: Sounds good. I added it

PR #152, opened 2021-04-15
Add a status function for integration into the statusline

mfussenegger:
That should also give some feedback and the people who really wanted could still override `dap.progress.report` to point to `vim.notify` or do something else with the information. (Though I don't consider the dap/progress.lua module API public/stable)


Shatur95:
Thank you! But it shows `stopped`, `paused`, etc. only after first stop for me.

mfussenegger:
> Thank you! But it shows stopped, paused, etc. only after first pause for me.

Could you elaborate or show a short video of what you see and explain what you'd expect to see?

Shatur95:
> Could you elaborate or show a short video of what you see and explain what you'd expect to see?

I integrated your plugin into my CMake helper [here](https://github.com/Shatur95/neovim-cmake-projects/blob/2eb08b8fb9715c70b4894daa0f3a0b6f0a52c072/lua/cmake/init.lua#L73). But `print(require('dap').status())` returns an empty string after starting session. I should call `pause` manually to enable any messages from `status` functions.

PR #159, opened 2021-04-18
Make variables in REPL expandable

mfussenegger:
Next step after this would be to extract some common `variables` widget that can be used in both the repl and the variables ui.

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -15,9 +15,6 @@ local history = {
 }
 
 
-local variables_ns = api.nvim_create_namespace('dap.repl.variables')
-
-
 M.commands = {
   continue = {'.continue', '.c'},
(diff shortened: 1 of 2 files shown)


theHamsta:
Can't the variables buffer could already be used in the REPL right? It only wants a buffer and a start line and indent to draw on.

mfussenegger:
> Can't the variables buffer could already be used in the REPL right? It only wants a buffer and a start line and indent to draw on. But of course the variables buffer could use the logic from the REPL. They are more and less interchangeable.

Yes approaching this from the other side would also be possible. I opted to take the REPL route because it had more constraints. The buffer can contain other information tahn the variables. That motivated the use of extmarks to assign arbitrary information on a line and it also had an impact on how to do the syntax highlights

PR #173, opened 2021-05-07
feat:  Add support for window split configuration

kconston:
Allow users to provide their own split command for opening a new window when using the integrated terminal.

A new property has been added to the configuration:  **windowSplit**.  If not provided, the default value of "belowright new" is used.

Example configurations:

### Open a new horizontal window below, with 20 height:
```
 require'dap'.configurations.python = {
    {
      type = 'python';
      request = 'launch';
      name = "Launch file";
      console = "integratedTerminal";
      windowSplit = 'botright 20new';
   }
}
```

### Open a new vertical window to the right, with 40 width:
```
 require'dap'.configurations.python = {
    {
      type = 'python'; 
      request = 'launch';
      name = "Launch file";
      console = "integratedTerminal";
      windowSplit = '40vsplit new';
   }
}
```

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -297,9 +297,24 @@ should be launched by nvim-dap:
 
 <
 
-`fallback` here can be replaced with the |dap-adapter| type to have type
-specific terminal configurations.
 
+If you're using the integrated terminal, you can also configure the command
+that is used to create a split window:
+
+
+>
+
+  lua << EOF
+  local dap = require('dap')
+  dap.defaults.fallback.terminal_win_cmd = '50vsplit new'
+  EOF
+<
+
+The `terminal_win_cmd` defaults to `belowright new`.
+
+
+`fallback` can be replaced with the |dap-adapter| type to have type
+specific terminal configurations.
 
 ==============================================================================
 API                                                  *dap-api*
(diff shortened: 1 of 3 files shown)


mfussenegger:
Thank you for the PR.

Sounds good to me to make this configurable but I think the `dap.configurations` should be kept for debug adapter configuration only and not be used for the configuration of the client behavior. Although it's unlikely, it could conflict with debug adapter options and I think it's also confusing to users if the same set of configurations can affect either the client or the debug adapter.

Instead it would be possible to utilize the `dap.defaults`. They're already used to configure the external terminal - among others. So it would fit well to configure the integrated terminal as well.
Maybe a `terminal_win_cmd` option?

See https://github.com/mfussenegger/nvim-dap/blob/f0c43dca363a52acf2b8e5a62363f2def40139f8/lua/dap/session.lua#L50 for an example of how it is used and see `:help dap-terminal` on how users can set the options.

kconston:
### Example  `dap.defaults` configuration: 
 `require'dap'.defaults.python.terminal_win_cmd = '50vsplit new'`

mfussenegger (commented):
Thansk again.

Added some docs and merged the PR.

PR #201, opened 2021-06-07
[WIP] Pass correct environment variables to termopen (integratedTerminal)

### LINKED ISSUE #191: Run command before launching
Hi, I was trying to use this plugin in a typescript plugin and wasn't able to get it running. I need to run a command (`yarn build`) before starting the application and attaching to it. Is this currently possible?

dchusovitin:
Fix #191 ([comment](https://github.com/mfussenegger/nvim-dap/issues/191#issuecomment-854442126))

I think nvim-dap should support only one type of env's in configuration, like this:

```lua
env = {
  ['PORT']  = '9001',
  ['GOOGLE_APPLICATION_CREDENTIALS'] = '${workspaceFolder}/.credentials/local.json'
}
```
And not:

```lua
env = {
  'PORT = 9001',
  'GOOGLE_APPLICATION_CREDENTIALS = "${workspaceFolder}/.credentials/local.json"'
}
```

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -77,6 +77,22 @@ function Session:run_in_terminal(request)
     clear_env = false;
     env = non_empty(body.env) and body.env or vim.empty_dict()
   }
+
+  if vim.tbl_count(opts.env) > 0 then
+    local term_env = {}
+    for k, v in pairs(opts.env) do
+      if type(k) == 'string' and type(v) == 'string' then
+        term_env[k] = v
+      elseif type(v) == 'string' then
+        local sep = v:find("=")
+        if sep ~= nil then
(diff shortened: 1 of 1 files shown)


mfussenegger (commented):
The debug adapter protocol specifies the type for `env`  in the `RunInTerminal` like this:

```
/**
   * Environment key-value pairs that are added to or removed from the default
   * environment.
   */
  env?: { [key: string]: string | null; }
```


So I don't think nvim-dap should be lenient and convert environment options that are in another format. If the debug adapter sends the environment variables in a different format, that is a debug adapter bug.


What options the debug adapter expects for the launch/attach configuration is theoretically unrelated - but ultimately also up to the debug adapter.

PR #217, opened 2021-06-19
Add more predefined variables

PatOConnor43:
### Summary
This PR adds more predefined variables like `${file}` and `${workspaceFolder}` which should match VS Code examples defined [here](https://code.visualstudio.com/docs/editor/variables-reference#_predefined-variables-examples).

### Motivation
I was trying to create a Go Delve configuration that used `${fileDirname}` based off these [docs](https://go.googlesource.com/vscode-go/+/HEAD/docs/debugging.md#debug-a-single-test-function). I don't know if you have any interest in supporting all of these but they were simple enough to construct so I figured I'd include as many as I could.

With this variable I can now construct a config that looks like this to debug all tests within a package:
``` lua
    {
      type = "go",
      name = "Debug package tests", -- configuration for debugging test files
      request = "launch",
      mode = "test",
      program = "${fileDirname}/...",
    },

```

Also, it looks like `${file}` in this repo is relative to the workspace but `${file}` in VS Code is absolute. I don't know if that was intentional or not, just an FYI. Thanks for taking the time to review :smile:

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -189,10 +189,15 @@ Things to note:
   they will be evaluated once the configuration is used.
 
 - Some variables are supported:
-
   - `${file}`: Active filename
+  - `${fileBasename}`: The current file's basename
+  - `${fileBasenameNoExtension}`: The current file's basename without extension
+  - `${fileDirname}`: The current file's dirname
+  - `${fileExtname}`: The current file's extension
+  - `${relativeFile}`: The current file relative to |getcwd()|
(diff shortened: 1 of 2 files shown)


mfussenegger:
Looks good to me. Could you also update the docs in the dap.txt?

PatOConnor43:
@mfussenegger I think this is ready to review again. Thanks!

mfussenegger:
Thanks again :+1: 
Merged the PR

PR #224, opened 2021-07-03
Accept adapter/debugee with different names in launch.json

weihanglo:
In a debugee configurations, the `type` field refers to an adapter
defined in `dap.adapters.<adapter-name>`. Current implementation has
an assumption that the `config.type`, which is the name of an adapter,
is the same as a debugee name. Therefore it is used for retrieving a
debugee configurations as a key of `dap.configurations[config.type]`.

The assumption works in most cases, since most adapter-debugee gets the
same name for which a simple "filetype" is often chosen, such as php or
python. One example breaks the rule is lldb adapter. There is a mismatch
between the adapter lldb and its multiple debugees like c++ and rust.

This commit simply groups filetypes by adapaters to fix it.

### DIFF
--- a/lua/dap/ext/vscode.lua
+++ b/lua/dap/ext/vscode.lua
@@ -16,15 +16,35 @@ function M.load_launchjs(path)
   end
   local contents = table.concat(lines, '\n')
   local data = vim.fn.json_decode(contents)
+
+  -- Groups filetypes of debugee configs by adapters if a config use the adapter.
+  -- { <adapter> = { <filetype> = boolean } }
+  local filetypes_of_adapters = {}
+  for filetype, configs in pairs(dap.configurations) do
+    for _, config in ipairs(configs) do
+      local filetypes = filetypes_of_adapters[config.type] or {}
+      filetypes_of_adapters[config.type] = filetypes
+      filetypes[filetype] = true
+    end
+  end
+
   assert(data.configurations, "launch.json must have a 'configurations' key")
   for _, config in ipairs(data.configurations) do
     assert(config.type, "Configuration in launch.json must have a 'type' key")
-    local configurations = dap.configurations[config.type]
-    if not configurations then
-      configurations = {}
+    local filetypes = filetypes_of_adapters[config.type]
+    if filetypes then
+      -- Add this config for filetypes containing configs with this adapter.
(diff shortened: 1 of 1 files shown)


weihanglo:
Also, thank you for creating such a great plugin 🎉 🎉

lua/dap/ext/vscode.lua
```diff
@@ -16,15 +16,29 @@ function M.load_launchjs(path)
...
+
   assert(data.configurations, "launch.json must have a 'configurations' key")
   for _, config in ipairs(data.configurations) do
     assert(config.type, "Configuration in launch.json must have a 'type' key")
-    local configurations = dap.configurations[config.type]
-    if not configurations then
-      configurations = {}
-      dap.configurations[config.type] = configurations
+    local debugees = debugees_of_adapter[config.type]
+    if debugees then
```
mfussenegger, line 35: Doesn't this mean that if `dap.configurations` is initially empty,  nothing would be added? 

Did you see the conversations in https://github.com/mfussenegger/nvim-dap/pull/156 ?



The debugger naming is also a bit confusing to me - the introduction in dap.txt refers to the application being debugged as debugee but in this PR it's used for the configuration indices

mfussenegger (commented):
Thank you for the PR, I  left some feedback

lua/dap/ext/vscode.lua
weihanglo: Oh, sorry. I didn't notice that PR. Also, It should be filetype not debugee itself, that's my fault.  

For the improvement, we can fallback to the old way (using adapter's name as filetype to add new debugee config) if there is no debugee config for that filetype. Doing it this way we need not to add any new mapping key between launch.json and nvim-dap.  What do you think?

mfussenegger: > For the improvement, we can fallback to the old way (using adapter's name as filetype to add new debugee config) if there is no debugee config for that filetype. Doing it this way we need not to add any new mapping key between launch.json and nvim-dap. What do you think?

Sounds good

weihanglo: Just fixed it. Let me know if anything need to improve 🙂

mfussenegger:
Thanks :+1: 

I merged the change but squashed the commits and reworded the commit message a bit.

Issue #232, opened 2021-07-09
Go (using delve directly) no output


y1rn:
as subject, i m using nvim-dap-ui,but no every log output in repl panel.

mfussenegger:
I'll need a more detailed problem description to investigate or help

y1rn:
![image](https://user-images.githubusercontent.com/70128797/125121676-d5fbfc00-e126-11eb-9fc4-f1b2cbadba33.png)
should be print something here, right?

mfussenegger:
Looks like `dlv dap` currently logs the output to its own stdout instead of routing it via the debug adapter protocol.


https://user-images.githubusercontent.com/38700/125174924-fd0b0a00-e1c8-11eb-8e6b-d7b614359ed4.mp4


Ideally they'd support the `runInTerminal` request, so you could have it launch your application in a dedicated terminal. Looks like that's not yet supported: https://github.com/golang/vscode-go/issues/124 - but there is active work in that direction going on: https://github.com/go-delve/delve/pull/2566

mfussenegger:
Actually there is something you can do, you could redirect the output of `dlv` to the repl:

```lua
dap.adapters.dlv_spawn = function(cb)
  local stdout = vim.loop.new_pipe(false)
  local handle
  local pid_or_err
  local port = 38697
  local opts = {
    stdio = {nil, stdout},
    args = {"dap", "-l", "127.0.0.1:" .. port},
    detached = true
  }
  handle, pid_or_err = vim.loop.spawn("dlv", opts, function(code)
    stdout:close()
    handle:close()
    if code ~= 0 then
      print('dlv exited with code', code)
    end
  end)
  assert(handle, 'Error running dlv: ' .. tostring(pid_or_err))
  stdout:read_start(function(err, chunk)
    assert(not err, err)
    if chunk then
      vim.schedule(function()
        --- You could adapt this and send `chunk` to somewhere else
        require('dap.repl').append(chunk)
      end)
    end
  end)
  -- Wait for delve to start
  vim.defer_fn(
    function()
      cb({type = "server", host = "127.0.0.1", port = port})
    end,
    100)
end
```



You'll have to change the `type` in your configurations to `dlv_spawn` to use this new adapter definition. For example:


```lua
dap.configurations.go = {
  {
    type = 'dlv_spawn',
    name = 'Launch dlv & file',
    request = 'launch',
    program = "${file}";
  },
}
```


[...]

PR #233, opened 2021-07-11
fix split repl lines issue

y1rn:
image

In this debug step by step case expect:
(
initial
1 2
 true
 0
 apple
)

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -274,6 +274,9 @@ function M.append(line, lnum)
     line = line:gsub('\r\n', '\n')
   end
   local lines = vim.split(line, '\n')
+  if #lines > 1 and lines[#lines] == '' then
+    table.remove(lines)
+  end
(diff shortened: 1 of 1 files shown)


mfussenegger:
I assume this is partially related to the `require('dap.repl').append(chunk)` from https://github.com/mfussenegger/nvim-dap/issues/232 ?

I think instead of trimming the last newline within the repl append code, this should rather be done on the call-site already.

You could wrap the chunk in an `vim.trim()` call.

y1rn:
i have been debug rust file, still has the same issue.

mfussenegger:
Ah, I see.

I've merged the change with https://github.com/mfussenegger/nvim-dap/commit/55f2737dcfb2f0d846623ae3996cf141bb692fab

y1rn:
i work with vim.trim() well, i'v amend my fork https://github.com/y1rn/nvim-dap/commit/d2bf364952eea328fcea33671e6654e59c736c26

PR #235, opened 2021-07-17
repl lines issue use vim.trim()

y1rn:


mfussenegger:
Does this fix any actual issue, or is this just cleaning up?

y1rn:
just clean up

PR #246, opened 2021-07-24
fix(ui): handle empty text in repl

rcarriga:
There's an error raised if a REPL command returns no text. For example assigning a variable in python.

```
Error executing vim.schedule lua callback: ...ache/nvim/site/pack/packer/start/nvim-dap/lua/dap/ui.lua:407: end_col value outside rang
```
This is more of a quick fix suggestion that I found while testing, not sure if this how you would want to handle this


mfussenegger:
The fix looks good to me.

PR #249, opened 2021-07-25
unwrap text with variable description too long

y1rn:
before:
before

after:
after


mfussenegger:
I kinda want to deprecate the variables-ui API and replace it with the widget API - so not sure if it makes sense to still merge changes to it.

For the time being you could create a `dap-variables.vim` in `.config/nvim/ftplugin` to customize the buffer options.

y1rn:
> I kinda want to deprecate the variables-ui API and replace it with the widget API - so not sure if it makes sense to still merge changes to it.
> 
> For the time being you could create a `dap-variables.vim` in `.config/nvim/ftplugin` to customize the buffer options.

i'v created dap-variables.vim with content ``` set nowrap```, it doesn't work.
``` echo &ft ``` with output ```dap-variables```
```:set ft=dap-variables``` and it works!

PR #270, opened 2021-08-12
Clarify how to use dap.repl custom_commands

amenonsen:
1. Make it clear in the example that one must require "dap.repl" before
   extending the dap.repl.commands table; otherwise the lazy-loaded repl
   won't see the custom commands in the extended table, even though the
   changes are visible to vim.inspect().

2. Also make it clear what you can or cannot hope to do by extending the
   commands table: you can add aliases for existing builtin commands
   (but not override them), you can hook up a new command to an existing
   dap function, or define an entirely new custom command as a function
   (and you only need to override what you want to change).

3. Earlier, the code would invoke the custom command with the whole of
   the command text, i.e., ".echo hello world". On the basis that it's
   inconvenient for every custom command to have to always strip the
   command name from its arguments, we now pass only "hello world".

   This change has the potential to break existing custom commands that
   expect the command name to be included in the function argument, but
   a Github code search found no custom_commands other than the '.echo'
   example copied from the docs, or commands that didn't take arguments.

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -610,29 +610,21 @@ repl.open({winopts}, {wincmd})                                 *dap.repl.open()*
           .rc or
           .reverse-continue   Same as |dap.reverse_continue|
 
-        You can customize the commands by overriding `dap.repl.commands`:
+        You can customize the builtin command names or define your own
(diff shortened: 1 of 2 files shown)


mfussenegger:
Thanks 👍

PR #279, opened 2021-08-22
fix error when decoded.command is nil

puuuuh:


mfussenegger:
Could you explain how this happens?

The `Response` of the debug adapter protocol defines the `command` as `string`. Without any optional or `null` marker. So the command should always be present, and if it isn't it is a bug in the debug adapter.

puuuuh:
It should be non-null, but vscode-lldb responses to launch request without this field. probably i should create issue in their repo instead

mfussenegger:
> probably i should create issue in their repo instead

Would be great if you could do that. I would like to avoid workarounds within nvim-dap.

PR #280, opened 2021-08-23
Use vim.notify

### LINKED ISSUE #111: Use vim.notify
We could use `vim.notify` to display notifications (https://github.com/neovim/neovim/pull/13843) to enable the use to handle different notification levels (info/warning/error) in their preferred format.

Shatur:
I replaced all `print` occurrences with `vim.notify`. I added `DAP` title, so users can display notifications nicer. For example, using [nvim-notify](https://github.com/rcarriga/nvim-notify) users can have nice popups.
Gitsigns [also use title](https://github.com/lewis6991/gitsigns.nvim/commit/4f0bf0d931da1d2b55f64a87a9fe11e39681194d).

Closes #111.


mfussenegger (commented):
I think the second arguments should use `vim.log.levels.XY`. See the stock implementation of `vim.notify`:

```
function vim.notify(msg, log_level, _opts)

  if log_level == vim.log.levels.ERROR then
    vim.api.nvim_err_writeln(msg)
  elseif log_level == vim.log.levels.WARN then
    vim.api.nvim_echo({{msg, 'WarningMsg'}}, true, {})
  else
    vim.api.nvim_echo({{msg}}, true, {})
  end
end
```

Using a string literal as log level won't be recognized.

I'm also not sure if it is justified to use error. Stuff like "No configuration selected" would be fine as info. Messages like "Restart not supported" could probably go as warning. I think "error" should be reserved for unexpected errors - like the when using the error object from the handler response.

Shatur:
> I think the second arguments should use vim.log.levels.XY. See the stock implementation of vim.notify:

Oh, right, completely missed it. For some reason using string works for `nvim-notify`. Will fix it.

> I'm also not sure if it is justified to use error. Stuff like "No configuration selected" would be fine as info. Messages like "Restart not supported" could probably go as warning. I think "error" should be reserved for unexpected errors - like the when using the error object from the handler response.

Makes sense.

> It may also be worth to extract the title = 'DAP' thing into a NOTIFY_OPTS constant or something?

Where should I store it?

Shatur:
> Messages like "Restart not supported" could probably go as warning.

Where this causes the function to stop, I left it as `INFO`. Usually "warning" means that something is working at least partially. Let me know if I should change it.

Shatur:
> Where should I store it?

@mfussenegger I meant in which script should I store this variable?

PR #281, opened 2021-08-25
feat: allow custom request handlers

entropitor:
Based off https://github.com/mfussenegger/nvim-dap/issues/136#issuecomment-876329841 I thought to give it another try for vscode-js-debug and I was able to get it working. The only thing I need is this request. While python uses an event to "announce" the sessions (https://github.com/mfussenegger/nvim-dap-python/pull/21/files), vscode-js-debug uses a request named `attachedChildSession`

PR #286, opened 2021-08-31
Allow adapters to define reverse request handlers

mfussenegger:
Alternative to https://github.com/mfussenegger/nvim-dap/pull/281

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -63,7 +63,7 @@ local function launch_external_terminal(terminal, args)
 end
 
 
-function Session:run_in_terminal(request)
+local function run_in_terminal(self, request)
   local body = request.arguments
   log.debug('run_in_terminal', body)
   local settings = dap().defaults[self.config.type]
@@ -583,17 +583,31 @@ function Session:handle_body(body)
         log.warn('No event handler for ', decoded)
       end
     end)
-  elseif decoded.type == 'request' and decoded.command == 'runInTerminal' then
-    self:run_in_terminal(decoded)
+  elseif decoded.type == 'request' then
+    local handler = self.handlers.reverse_requests[decoded.command]
+    if handler then
+      handler(self, decoded)
+    end
+    log.warn('No handler for reverse request', decoded)
   else
     log.warn('Received unexpected message', decoded)
   end
 end
 
 
-local function session_defaults(opts)
+local default_reverse_request_handlers = {
+  runInTerminal = run_in_terminal
+}
+
+
+local function session_defaults(adapter, opts)
   local handlers = {}
   handlers.after = opts.after
+  handlers.reverse_requests = vim.tbl_extend(
+    'error',
+    default_reverse_request_handlers,
+    adapter.reverse_request_handlers or {}
+  )
   return {
     handlers = handlers;
     message_callbacks = {};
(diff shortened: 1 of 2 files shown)


lua/dap/session.lua
```diff
@@ -560,17 +560,31 @@ function Session:handle_body(body)
...
+local default_reverse_request_handlers = {
+  runInTerminal = run_in_terminal
+}
+
+
+local function session_defaults(adapter, opts)
   local handlers = {}
   handlers.after = opts.after
+  handlers.reverse_requests = vim.tbl_extend(
+    'keep',
```
entropitor, line 584: so this means that the default wins? And if the adapter has it as well, it will not be overridden?

Should it maybe be `'error'` such that if the adapter overrides, an error is thrown instead of it being silently swallowed? This will hopefully make it a bit more clear that something is off (after an upgrade). Potentially we could even catch the error here and throw it with a clear error message that the adapter is overriding a handler it cannot override (because there is now a default)

lua/dap/session.lua
```diff
@@ -560,17 +560,31 @@ function Session:handle_body(body)
...
 end
 
 
-local function session_defaults(opts)
+local default_reverse_request_handlers = {
+  runInTerminal = run_in_terminal
+}
+
+
+local function session_defaults(adapter, opts)
```
entropitor, line 580: Having the adapter here will be useful as I think another nice-to-have for vscode-js-debug would be to have a custom onExit for the session. This change should make that trivial to implement.

entropitor (approved):
Looks good to me. I think it's definitely better than mine proposal (so I'll close my PR). The only question is whether we should go for `keep`, `error` or `force` I guess? I'll leave the decision to you. You probably know best here.

I guess the keep is also a safe default to begin with and it can more easily be reverted than starting with force so I think it might be a smart start and it would at least enable vscode-js-debug which is the most important thing at this moment.

lua/dap/session.lua
mfussenegger: Good point, I'll change it to error

lua/dap/session.lua
mfussenegger: Could you elaborate why that would be useful - and why the existing `dap.listeners` for the exit/terminated events wouldn't be sufficient?

entropitor: The problem is that for javascript, you need to run a command which starts the server and outputs a port and this is the port you need to use in the adapter above. But at that point there is no session yet, so you cannot attach the onexit (to stop the original process) on the session

https://github.com/entropitor/dotfiles/blob/06aa5ea67fad9be8e828ec0cafbada7b186df82b/tag-vim/config/nvim/lua/entropitor/dap.lua#L9
https://github.com/entropitor/dotfiles/blob/06aa5ea67fad9be8e828ec0cafbada7b186df82b/tag-vim/config/nvim/lua/entropitor/dap/dap-js.lua#L88

mfussenegger: You can register listeners at any time and registering them doesn't require a session.
The `connect` method could do something like this:


```lua
dap.listeners.before['event_terminated']['your-unique-plugin-id'] = function(session, body)
  -- stop process here

  dap.listeners.before['event_terminated']['your-unique-plugin-id'] = nil
end
```

mfussenegger:
Pushed to master via https://github.com/mfussenegger/nvim-dap/commit/409fd0cfbaa39e52e02c06eeae5456cadb0b6748

Issue #289, opened 2021-09-08
Feature Request: Ability to highlight conditional breakpoints separately from standard breakpoints


knpwrs:
### Problem Statement

It would be great if I could highlight breakpoints in red and conditional breakpoints in yellow.

danielnehrig:
how would you like this to be implemented @mfussenegger
i'd do a  "DapBreakpointConditional" sign define and add that when the conditional hits

is that ok ?

mfussenegger:
https://github.com/mfussenegger/nvim-dap/pull/312 got merged

PR #312, opened 2021-09-23
feat(breakpoint): add conditional breakpoint sign define

danielnehrig:
addressing feature #289

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -228,7 +228,7 @@ call it once.
 SIGNS CONFIGURATION
 
 
-nvim-dap uses four signs:
+nvim-dap uses five signs:
 
 - `DapBreakpoint` which defaults to `B` for breakpoints
 - `DapBreakpointCondition` which defaults to `C` for conditional breakpoints
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -231,6 +231,7 @@ SIGNS CONFIGURATION
 nvim-dap uses four signs:
 
 - `DapBreakpoint` which defaults to `B` for breakpoints
+- `DapBreakpointCondition` which defaults to `C` for conditional breakpoints
 - `DapLogPoint` which defaults to `L` and is for log-points
 - `DapStopped` which defaults to `→` and is used to indicate the position where
   the debugee is stopped.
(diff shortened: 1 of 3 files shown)


lua/dap/breakpoints.lua
```diff
@@ -68,10 +68,17 @@ function M.toggle(opts, bufnr, lnum)
   if M.remove(bufnr, lnum) and not opts.replace then
     return
   end
+  local sign_name = 'DapBreakpoint'
+  if (non_empty(opts.condition)) then
+    sign_name = 'DapBreakpointCondition'
+  end
+  if (non_empty(opts.log_message)) then
+    sign_name = 'DapLogPoint'
+  end
```
danielnehrig, line 77: is that ok like this
or should the be a check to ensure that not both properties of the table are non empty

mfussenegger: The protocol specification allows both to be set - so this is fine. Or at least doesn't state explicitly that they shouldn't. So no reason to add artificial restrictions.

doc/dap.txt
```diff
@@ -231,6 +231,7 @@ SIGNS CONFIGURATION
 nvim-dap uses four signs:
```
mfussenegger, line 231:
```suggestion
nvim-dap uses five signs:
```

mfussenegger:
I'm wondering if it would make sense to change the approach altogether and add multiple signs.  One for the breakpoint,  one for the log, one for the condition.  Depending on the sign column setting you see only one or all

danielnehrig:
@mfussenegger do you have an idea why the ci failed
```
Run curl -OL https://raw.githubusercontent.com/norcalli/bot-ci/master/scripts/github-actions-setup.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  2474  100  2474    0     0   172k      0 --:--:-- --:--:-- --:--:--  172k

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Error: Process completed with exit code 2.
```

theHamsta:
@danielnehrig the Neovim nightly is not available today. You can try again tomorrow. Actually nvim-dap could also be tested on 0.5.1 (just like nvim-treesitter which also has failing CIs today)

danielnehrig:
@theHamsta thanks for the info

PR #341, opened 2021-10-23
feat: enable jumping to stack frames of other threads

theHamsta:
At the moment, this would only be enabled in REPL when `allThreadsStopped`.

### DIFF
--- a/lua/dap/entity.lua
+++ b/lua/dap/entity.lua
@@ -134,5 +134,20 @@ function frames.render_item(frame)
 end
 
 
+local threads = {}
+M.threads = threads
+
+function threads.render_item(thread)
+  local session = require('dap').session()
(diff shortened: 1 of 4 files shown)


mfussenegger (commented):
Sorry for not getting back to this earlier.

I've had a thread widget prototype when I was working on the widget system that allowed to list the threads and have several actions on them e.g. to pause them and so on.

I cut that away at some point to keep the initial changes smaller and didn't follow up on bringing it back.

If it's okay with you I'd cherry pick some of the stuff in this PR (with attribution of course) after I've integrated some of the changes from https://github.com/mfussenegger/nvim-dap/pull/257

mfussenegger:
Closing this as https://github.com/mfussenegger/nvim-dap/pull/498 and https://github.com/mfussenegger/nvim-dap/pull/500 should cover it.

Issue #291, opened 2021-09-08
When entering insert mode again from normal mode in repl , the backspace can't delete character.


pittcat:
### Debug adapter definition and debug configuration

[nvim-dap-python](https://github.com/mfussenegger/nvim-dap-python)

### Debug adapter version

latest

### Steps to Reproduce

1. Input any character in repl buffer
2. Press <Esc> to leave insert mode
3. Enter insert mode again
4. Press <Backspace> to delete character

### Expected Result

Backspace can delete character when entering insert mode again by some workaround.

### Actual Result

https://asciinema.org/a/oQgELi5G8rkCvUfqLNnSHnuty

mfussenegger:
As far as I'm aware this is is a vim/neovim issue with the prompt buftype. I attempted workarounds, but so far failed to come up with anything that is reliable.

This needs to be fixed upstream.

wookayin:
@mfussenegger Do you have any pointer to the upstream issue? Thanks!

mfussenegger:
@wookayin I think https://github.com/neovim/neovim/pull/15881 might be related, but I didn't verify.

pfeiferj:
@pittcat I put my workaround for this in #119 ([comment](https://github.com/mfussenegger/nvim-dap/issues/119#issuecomment-976046944)) not realizing there was an open issue for it. I also created #357 to potentially resolve this but not sure if it's something @mfussenegger would really want to merge in because it's not very clean.

pfeiferj:
An improvement to the previously mentioned backspace workaround that doesn't switch between normal and insert modes:

```lua
function PromptBackspace()
  -- Allows backspacing through previously set text when in a prompt.
  --
  -- Note 1: Insert mode cursor is after (+1) the column as opposed to in normal mode it would be on (+0) the column.
  -- Note 2: nvim_win_[get|set]_cursor is (1, 0) indexed for (line, column) while nvim_buf_[get|set]_[lines|text] is 0 indexed for both.

  local currentCursor = vim.api.nvim_win_get_cursor(0)
  local currentLineNumber = currentCursor[1]
  local currentColumnNumber = currentCursor[2]
  local currentLine = vim.api.nvim_buf_get_lines(0, currentLineNumber-1, currentLineNumber, false)[1]
  local _, endPromptPrefix = string.find(currentLine, "> ")
 
  if endPromptPrefix == nil then
    endPromptPrefix = 0
  end

[...]

PR #357, opened 2021-11-23
Add remap for backspace in repl

pfeiferj:
Add a remap for backspace in the repl to allow previously written text to be backspaced through.

This should resolve the issue in #119 and #291 without out any major side effects that I'm aware of.

I realize this is a pretty hacky workaround so I definitely won't be offended if it gets rejected.

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -285,6 +285,10 @@ function M.prompt_backspace()
   local currentLine = vim.api.nvim_buf_get_lines(0, currentLineNumber-1, currentLineNumber, false)[1]
   local _, endPromptPrefix = string.find(currentLine, "> ")
 
+  if endPromptPrefix == nil then
+    endPromptPrefix = 0
+  end
+
   if (currentColumnNumber-1) ~= endPromptPrefix then
     vim.api.nvim_buf_set_text(0, currentLineNumber-1, currentColumnNumber-2, currentLineNumber-1, currentColumnNumber-1, {""})
     vim.api.nvim_win_set_cursor(0, {currentLineNumber, currentColumnNumber-2})
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -279,14 +279,14 @@ function M.prompt_backspace()
   -- Note 1: Insert mode cursor is after (+1) the column as opposed to in normal mode it would be on (+0) the column.
   -- Note 2: nvim_win_[get|set]_cursor is (1, 0) indexed for (line, column) while nvim_buf_[get|set]_[lines|text] is 0 indexed for both.
 
-  local currentCursor = vim.api.nvim_win_get_cursor(0)
-  local currentLineNumber = currentCursor[1]
-  local currentColumnNumber = currentCursor[2]
-  local promptLength = string.len(vim.fn['prompt_getprompt']('%'));
+  local current_cursor = vim.api.nvim_win_get_cursor(0)
+  local current_line_number = current_cursor[1]
+  local current_column_number = current_cursor[2]
+  local prompt_length = string.len(vim.fn['prompt_getprompt']('%'));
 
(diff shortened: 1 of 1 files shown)


pfeiferj:
One thing to note is that in the key mapping I'm doing `<esc>:` instead of `<Cmd>` on purpose. I tried doing a `stopinsert` from inside of the prompt_backspace function but it resulted in some weird behavior surrounding how it exits and re-enters insert mode.

bew:
Can't you just edit/set the line from lua (get the line, modify it, set it, and take care of the cursor position), without having to go through normal mode to do this?

(note: didn't tried)

pfeiferj:
> Can't you just edit/set the line from lua (get the line, modify it, set it, and take care of the cursor position), without having to go through normal mode to do this?
> 
> (note: didn't tried) 

I can try that this evening when I get off work. I suspect there will be some weirdness around getting text in insert mode that will either not get all of the text or cause you to leave insert mode. I started out writing in vimscript and had some issues that caused any text you entered in the current insert mode to disappear, but I'm not sure what exactly caused that. If you could just edit the line in lua that would be much nicer though.

pfeiferj:
>Can't you just edit/set the line from lua (get the line, modify it, set it, and take care of the cursor position), without having to go through normal mode to do this?
>
>(note: didn't tried)

So I started out trying vims set_line function which actually ended up triggering a submission so didn't work out. Fortunately the neovim api has nvim_buf_set_text which can replace a partial line and does work. Only issue I ran into is that the neovim api is a little inconsistent on if the line/column is 0 or 1 based and if you get the cursor position in insert mode the cursor is after the line meaning you're +1 from where you would be in normal mode. It still results in more elegant code though so i'll update the pr.

mfussenegger (commented):
I'm a bit mixed and undecided on this.

I'd prefer if this got fixed in neovim itself - so that not every plugin has to come up with a workaround like this.

On the other hand, this is pretty isolated and if it works, why not.

I'll leave it up for a while longer to decide.

pfeiferj:
Yeah, I definitely agree that this one is a judgement call. For what it's worth, I'm personally running this with an autocommand that applies it to all prompts anyways, so not merging it wouldn't affect me. The way I see it there's a few valid options here.

* Don't merge it and hope for an upstream fix. I might try looking into fixing it upstream myself and submitting a pr, but that also might be more work than I want to do. I'd have to see if it's a relatively easy fix before putting time into it.
*  Don't merge it but provide a faq entry/documentation that points to the upstream issue but also describes how to setup the workaround. This still gives visibility to the issue, but some people may not see the workaround and still face the issue.
* Merge it and risk lowering visibility into the upstream issue. This would provide a better user experience for this plugin than without the patch though.

Let me know if you would prefer the faq/documentation approach and I'll update the pr accordingly.

lua/dap/repl.lua
```diff
@@ -272,6 +273,22 @@ function M.on_down()
...
+function M.prompt_backspace()
+  -- Allows backspacing through previously set text when in a prompt.
+  --
+  -- Note 1: Insert mode cursor is after (+1) the column as opposed to in normal mode it would be on (+0) the column.
+  -- Note 2: nvim_win_[get|set]_cursor is (1, 0) indexed for (line, column) while nvim_buf_[get|set]_[lines|text] is 0 indexed for both.
+
+  local currentCursor = vim.api.nvim_win_get_cursor(0)
+  local currentLineNumber = currentCursor[1]
+  local currentColumnNumber = currentCursor[2]
+  local promptLength = string.len(vim.fn['prompt_getprompt']('%'));
```
theHamsta, line 285: You could use snake case to be more consistent with the rest of the code base

pfeiferj: good point, I'll get an update together here in a little bit

pfeiferj:
Closing as this is resolved by https://github.com/neovim/neovim/pull/16421 also see #291

PR #376, opened 2021-12-10
feat: add a sec_to_wait config value to delay warning

ckipp01:
For more context you can look at my comment here: https://github.com/mfussenegger/nvim-dap/discussions/314#discussioncomment-1782131

I wasn't sure if you wanted this explicitly stated int he docs since this might be sort of niche. Let me know your thoughts!

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -847,8 +847,12 @@ function Session:initialize(config, adapter)
     end)
   end)
   local sec_to_ms = 1000
+  local sec_to_wait = 4
+  if adapter.options and adapter.options.initialize_timeout_sec then
+    sec_to_wait = adapter.options.initialize_timeout_sec
+  end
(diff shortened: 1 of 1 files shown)


lua/dap/session.lua
```diff
@@ -847,8 +847,9 @@ function Session:initialize(config, adapter)
     end)
   end)
   local sec_to_ms = 1000
+  local sec_to_wait = config.sec_to_wait or 4
```
mfussenegger, line 850: `config` is reserved for debug adapter configuration. Client configuration should either go into the `defaults` table or in this case I think I'd prefer being it part of the `adapter` definition.  For example `adapter.options.initialize_timeout_sec`

ckipp01: Sure, I changed this to then be part of the adapter options.

PR #403, opened 2022-01-11
Split newlines in output from REPL commands

gpanders:
For example, when using LLDB one can use the following command in the
REPL to create the following output:

    dap> `p $rsp
    (lldb) p $rsp\n(unsigned long) $0 = 140732920752800\n

Note that the returned text contains newline characters (\n), but it is
all on a single line.

This commit splits the output on these newline characters so that the
output instead looks like this:

    dap> `p $rsp
    (lldb) p $rsp
    (unsigned long) $0 = 140732920752800

### DIFF
--- a/lua/dap/ui.lua
+++ b/lua/dap/ui.lua
@@ -437,8 +437,8 @@ function M.layer(buf)
             .. vim.inspect(xs)
           ))
         end
-        text = text:gsub('\n', '\\n')
-        api.nvim_buf_set_lines(buf, i, i + 1, true, {text})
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks for the PR - I went with a slightly different approach for now: https://github.com/mfussenegger/nvim-dap/pull/410
Not sure if I'll keep it at that. I noticed that I often prefer the newline's being preserved in variables instead of getting expanded.

Maybe I'll add a toggle option at some point or something like that.

Issue #346, opened 2021-11-09
Failed to launch - Golang


ITHDevelopment:
### Debug adapter definition and debug configuration

Installed the adapter via : 
`go get -u github.com/go-delve/delve/cmd/dlv`

Configuration via nvim-dap-go:
`require('dap-go').setup()`

OS: windows 10
shell: powershell
Nvim v0.5.1

### Debug adapter version

1.7.2

### Steps to Reproduce

1. Create a golang project
2. Make folders in the project root, and make go modules within them
3. Import the external and local modules correctly, also set up go.mod files correctly
4. Run: `go run . ` and it works fine

1. In file .cmd/main.go
2. Set a breakpoint
3. Start debug session

### Expected Result

Start the debug session correctly, also build and run the code without problems just like using the `go run .` command

### Actual Result

Error on launch: Failed to launch
(It seems to not be detecting the modules correctly)
![Screenshot (7)](https://user-images.githubusercontent.com/71849483/140842009-cef73e27-ccf2-451c-8708-3df2a6c3ebec.png)

mfussenegger:
I think this might be a problem with the default configurations in nvim-dap-go

See https://github.com/mfussenegger/nvim-dap/issues/200#issuecomment-877566539 and https://github.com/mfussenegger/nvim-dap/issues/200#issuecomment-877798483

This is what dap-go sets: 

https://github.com/leoluz/nvim-dap-go/blob/624a8b610083f8206cd3988cefb9549b8ffe2799/lua/dap-go.lua#L71

Cc @leoluz

leoluz:
@ITHDevelopment Im not sure I understand your step 2 "Make modules inside other folders". Can you clarify what you mean by that pls?

BTW, I have a testing project that loads many different dependencies via Go Modules and I am able to run the main and debug it without issues. You can check the project here: https://github.com/leoluz/ingress-test

In order to start the debug you have to run `:lua require'dap'.continue()` (or whatever key binding you have mapped to it). You will be presented with the following options:

```
Configuration:                                                                                                                                         
1: Debug
2: Debug test
3: Debug test (go.mod)
Type number and <Enter> or click with the mouse (q or empty cancels):
```

Then you select option 1 as you are trying to debug your main module.

That should be it.

ITHDevelopment:
Okay I see the problem now, I had the go.mod and go.sum files inside cmd folder. Moving the files outside the folder resolve the initial problem, but now the debugger don't recognize package main functions.
[...]

PR #433, opened 2022-02-11
fix: Schedule vim notify

JRasmusBm:
**Why** is the change needed?

To prevent the following error:

```
Error executing luv callback:
vim.lua:431: E5560: nvim_err_writeln must not be called in a lua loop callback
stack traceback:
        [C]: in function 'nvim_err_writeln'
        vim.lua:431: in function 'notify'
        .../dotfiles/vim/pack/bundle/opt/nvim-dap/lua/dap/utils.lua:95: in function 'notify'
        ...bm/.config/nvim/pack/bundle/opt/nvim-dap/lua/dap/rpc.lua:69: in function <...bm/.config/nvim/pack/bundle/opt/nvim-dap/lua/dap
/rpc.lua:67>
```

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -367,7 +367,7 @@ function M.terminate(terminate_opts, disconnect_opts, cb)
       local opts = terminate_opts or vim.empty_dict()
       session:request('terminate', opts, function(err)
         assert(not err, vim.inspect(err))
-        vim.notify('Session terminated')
+        vim.schedule(function () vim.notify('Session terminated') end)
         if cb then
           cb()
         end
@@ -377,7 +377,7 @@ function M.terminate(terminate_opts, disconnect_opts, cb)
       M.disconnect(opts, cb)
     end
   else
-    vim.notify('No active session')
+    vim.schedule(function () vim.notify('No active session') end)
     if cb then
       cb()
     end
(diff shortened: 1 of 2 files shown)


mfussenegger (commented):
Some of the calls you changed are not happening inside a lua callback and aren't necessary.

Could you explain how you produced the error?

JRasmusBm:
@mfussenegger What a quick response! :smile: Thank you so much for all the effort you put into the community!

I am integrating with [Debugpy.nvim](https://github.com/HiPhish/debugpy.nvim) with [vim-test](https://github.com/vim-test/vim-test) so that I can debug unit tests in a docker container. I need to run a vim.defer_fn on the `Debugpy` call to make sure the debugee has time to start.

I first submitted the linked PR to [Debugpy.nvim](https://github.com/HiPhish/debugpy.nvim), but thought that others might try `dap.run` inside of lua callbacks, and that it is thus more appropriate here. Like with all PRs you are free to accept or reject, I just thought I would push what was useful for me back upstream. :smile:  

My related dotfile changes: [dotfiles #346 - Set up debugging for python](https://github.com/JRasmusBm/dotfiles/pull/346)

mfussenegger:
I think if you run `dap.run` itself in a luv callback it is better to wrap the whole `dap.run` call there. Otherwise I'd have to make sure to wrap *all* calls that are not safe going forward and that's not really something I want to test and pay attention to.

PR #456, opened 2022-03-03
Added sourceModified flag to payload to make Unity debugger work

BliznyukNM:
Without this flag VS code plugin was crashing on adding breakpoints.
I am not sure about making it always false though. I have no idea how to detect that.

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -484,6 +484,7 @@ do
           path = path;
           name = vim.fn.fnamemodify(path, ':t')
         };
+        sourceModified = false;
         breakpoints = buf_bps;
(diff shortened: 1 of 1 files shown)


mfussenegger:
`sourceModified` is supposed to be optional according to the debug adapter protocol specification. If the unity debugger crashes without this flag then that's almost certainly a bug in the debug adapter.


I'm pretty sure that just setting it to `false` would be wrong - it would probably need to track if changes happened between `setBreakpoints` requests. The usual approach in neovim to do something like that is via the `b:changedtick`

mfussenegger:
Looks like other clients also set it to `false` without tracking if the source is changed, so I assume it at least wouldn't break anything so I'm merging this after all.

PR #457, opened 2022-03-04
Ensure session is terminated before starting a new one

mfussenegger:


mfussenegger:
@David-Kunz fyi - this may be interesting to you. `dap.run` now automatically terminates a session if it's still active before starting a new one.

David-Kunz:
Hi @mfussenegger ,

Awesome, I will test it asap! Thank you!

David-Kunz:
Hi @mfussenegger ,

Unfortunately it doesn't work for me. I notice a longer delay when initiating a new debugging session, but the corresponding processes are not terminated.

Best regards,
David

Issue #475, opened 2022-03-12
vscode-cpp does not send terminated / exited event


Electrobub:
### Debug adapter definition and debug configuration

Download and set up vscode-cpptools following the details on the wiki: https://github.com/mfussenegger/nvim-dap/wiki/C-C---Rust-(gdb-via--vscode-cpptools)

Config:

```
dap.adapters.cppdbg = {
  id = 'cppdbg',
  type = 'executable',
  command = vim.fn.stdpath "data" .. '/dap/vscode-cpptools/OpenDebugAD7',
  -- command = vim.fn.stdpath "data" .. '/dap/vscode-server/extension/debugAdapters/bin/OpenDebugAD7',
}
dap.configurations.cpp = {
  {
    name = "Launch file",
    type = "cppdbg",
    request = "launch",
    program = function()
      return vim.fn.input('Path to executable: ', vim.fn.getcwd() .. '/', 'file')
    end,
    cwd = '${workspaceFolder}',
    stopOnEntry = true,
    -- runInTerminal = "false",
    -- terminal = "integrated",
    -- externalConsole = false,
    setupCommands = {  
        { 
            text = '-enable-pretty-printing',
            description =  'enable pretty printing',
            ignoreFailures = false 
        },
    },
  },
}
```

### Debug adapter version

1.8.4

### Steps to Reproduce

Initialise debugging (in conjunction with `nvim-dap-ui` in the event listener - it is clear to see that the event is sent)

Set breakpoints -> Debug, stepover, stepin, continue -> this works :) 

However, when I disconnect / terminate, the `terminated` or `exited` event is not sent

### Expected Result

The `terminated` or `exited` event triggering when the DAP session is stopped

### Actual Result

The last section of the log does not show the `terminated` or `exited` events (it just states the `disconnect` command)

```
...
...
[ DEBUG ] 2022-03-12T13:25:21Z+0100 ] ...nvim/site/pack/packer/start/nvim-dap/lua/dap/session.lua:563 ]	{
  body = {
    category = "telemetry",
    data = {
      ["VS.Diagnostics.Debugger.AdapterId"] = "cppdbg",
      ["VS.Diagnostics.Debugger.DebugCompleted.BreakCounter"] = 2,
      ["VS.Diagnostics.Debugger.EngineVersion"] = "17.1.20204.1",
      ["VS.Diagnostics.Debugger.HostVersion"] = "17.1.20204.1",
[...]

Issue #469, opened 2022-03-10
[php] terminating session is incomplete


JSteitz:
### Debug adapter definition and debug configuration

```lua
dap.adapters.php = {
  type = "executable",
  command = "node",
  args = { "--no-deprecation", vim.fn.stdpath("data") .. "/dap/vscode-php-debug/out/phpDebug.js" },
}

dap.configurations.php = {
  {
    name = "Launch Xdebug",
    type = "php",
    request = "launch",
    port = 9003,
    stopOnEntry = false,
    pathMappings = {
      ...
    },
    xdebugSettings = {
      max_children = 1024,
      max_data = 1000000,
      show_hidden = 1
    }
  }
}
```

### Steps to Reproduce

1. Open any php project to debug
2. Run continue()
3. Run terminate()

### Expected Result

Session is terminated and the terminate event is fired.

### Actual Result

It looks like the session is not completely terminated and the terminate event is not fired.

What I see, is that the node process is terminated, but when I try to run `continue()` again, I'm prompted with an existing session in which I have to decide what to do (terminate, disconnect, etc…)
Picking terminate or disconnect results in a `'disconnect' timed out...` message.

The only workarounds for me currently are:
1. to call `close()` after `terminate()` but this does not fire required events
2. or I close neovim and start from scratch (not really practical)

```
[ DEBUG ] 2022-03-10T16:19:54Z+0100 ] ...nvim/site/pack/packer/start/nvim-dap/lua/dap/session.lua:715 ]	"Spawning debug adapter"	{
  args = { "--no-deprecation", "/home/jura/.local/share/nvim/dap/vscode-php-debug/out/phpDebug.js" },
  command = "node",
  type = "executable"
}
[ DEBUG ] 2022-03-10T16:19:54Z+0100 ] ...nvim/site/pack/packer/start/nvim-dap/lua/dap/session.lua:887 ]	"request"	{
  arguments = {
    adapterID = "nvim-dap",
    clientId = "neovim",
    clientname = "neovim",
    columnsStartAt1 = true,
    linesStartAt1 = true,
    locale = "en_US.UTF-8",
    pathFormat = "path",
    supportsRunInTerminalRequest = true,
    supportsVariableType = true
  },
  command = "initialize",
  seq = 0,
  type = "request"
}
[ DEBUG ] 2022-03-10T16:19:54Z+0100 ] ...nvim/site/pack/packer/start/nvim-dap/lua/dap/session.lua:563 ]	{
[...]

PR #481, opened 2022-03-16
Close session on disconnect response

mfussenegger:
Closes https://github.com/mfussenegger/nvim-dap/issues/475
Closes https://github.com/mfussenegger/nvim-dap/issues/469

A consequence of this change is that even for the debug adapters that do
send a terminated event after receiving a disconnect request (like
java-debug), this event might not be received anymore because the
session is closed after the disconnect response.

Extensions that want to run some logic if a debug session ends
should listen to both `event_terminated` and `disconnect`.

@theHamsta @rcarriga fyi I think this might affect your plugins

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -385,9 +385,6 @@ end
 
 function Session:event_terminated()
   self:close()
-  if self.handlers.after then
-    self.handlers.after()
-  end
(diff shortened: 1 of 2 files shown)


rcarriga:
Thanks for the heads up!

PR #520, opened 2022-04-15
Define some commands by default

mfussenegger:


plugin/dap.lua
```diff
@@ -0,0 +1,24 @@
...
+    nargs = 1,
+    complete = function()
+      return vim.tbl_keys(require('dap.log').levels)
+    end
+  }
+)
+cmd('DapShowLog', 'split | e ' .. vim.fn.stdpath('cache') .. '/dap.log | normal! G', {})
+cmd('DapContinue', function() require('dap').continue() end, { nargs = 0 })
+cmd('DapToggleBreakpoint', function() require('dap').toggle_breakpoint() end, { nargs = 0 })
+cmd('DapOpenRepl', function() require('dap').toggle_breakpoint() end, { nargs = 0 })
```
bew, line 21: typo here? action should be
```suggestion
cmd('DapOpenRepl', function() require('dap').repl.open() end, { nargs = 0 })
```

dhruvmanila:
Would it be better if instead of having different commands for every action, we can have one single command `Dap` which takes a single argument which is a function name like so `:Dap continue`. I did this in my [dotfiles](https://github.com/dhruvmanila/dotfiles/blob/775bd87fdd63011f8069f3a8a0ff69afb2c4d562/config/nvim/plugin/commands.lua#L31-L62).

mfussenegger:
> Would it be better if instead of having different commands for every action, we can have one single command `Dap` which takes a single argument which is a function name like so `:Dap continue`. I did this in my [dotfiles](https://github.com/dhruvmanila/dotfiles/blob/775bd87fdd63011f8069f3a8a0ff69afb2c4d562/config/nvim/plugin/commands.lua#L31-L62).

Why would that be better?
From a user perspective I think dedicated commands makes it easier to discover the various commands due to the completion behavior of the command line. Though I guess preferences vary for this.

From an implementation detail it would mean giving up on some of the things you get from from the command system - like nargs/completion. You'd have to implement one completion function that can handle all the sub-commands.

plugin/dap.lua
mfussenegger: Thanks for the pointer.

dhruvmanila:
Right, understood. This makes more sense 👍

PR #535, opened 2022-04-23
feat: custom console

rcarriga:
Hey there :smile: So I'm trying to implement https://github.com/rcarriga/nvim-dap-ui/issues/88 but the way the integrated console works right now makes it impossible to hide the buffer without shutting down the process. The current method also requires a window to be open when starting. I was originally going to make a PR switching to using `jobstart` and `nvim_create_term` to allow background operation but then I also needed control over the buffer etc so thought it'd make more sense to just make it totally customisable.

This allows the user to set an `open_console` setting (I'm not sure if I've placed this setting in the correct place) which will take the request body and return a PID. My use case will look something like this (and will be set by dapui)
```lua
    dap.defaults.open_console = function(body)
      local jobid

      local chan = api.nvim_open_term(console_buf, {
        on_input = function(_, _, _, data)
          pcall(api.nvim_chan_send, jobid, data)
        end,
      })

      local opts = {
        clear_env = false,
        env = next(body.env or {}) and body.env or vim.empty_dict(),
        cwd = (body.cwd and body.cwd ~= "") and body.cwd or nil,
        height = height,
        width = width,
        pty = true,
        on_stdout = function(_, data)
          data = table.concat(data, "\r\n")
          api.nvim_chan_send(chan, data)
        end,
      }

      jobid = vim.fn.jobstart(body.args, opts)

      return vim.fn.jobpid(jobid)
    end
  end,

```

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -54,6 +54,7 @@ M.defaults = setmetatable(
       focus_terminal = false;
       auto_continue_if_many_stopped = true;
     },
+    open_console = false;
   },
   {
     __index = function(tbl, key)
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -97,26 +97,9 @@ local function launch_external_terminal(terminal, args)
   return handle, pid_or_err
 end
 
-
-local function run_in_terminal(self, request)
-  local body = request.arguments
-  log.debug('run_in_terminal', body)
-  local settings = dap().defaults[self.config.type]
-  if body.kind == 'external' or (settings.force_external_terminal and settings.external_terminal) then
-    local terminal = settings.external_terminal
-    if not terminal then
-      utils.notify('Requested external terminal, but none configured. Fallback to integratedTerminal', vim.log.levels.WARN)
-    else
-      local handle, pid = launch_external_terminal(terminal, body.args)
-      if not handle then
-        utils.notify('Could not launch terminal ' .. terminal.command, vim.log.levels.ERROR)
-      end
-      self:response(request, {
-        success = handle ~= nil;
-        body = { processId = pid; };
-      })
-      return
-    end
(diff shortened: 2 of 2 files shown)


mfussenegger:
Wouldn't it be enough to use the existing `terminal_win_cmd` option to take control of the window/buffer creation?

If the users don't want the window at all they can usually set the debug adapter to run in console mode instead of in integratedTerminal mode.


I'm a bit reluctant to allow customizing the integrated terminal behavior more given how core it is to the execution of the debug adapter. It would make triaging issues more difficult.

rcarriga:
> Wouldn't it be enough to use the existing terminal_win_cmd option to take control of the window/buffer creation?

Unfortunately not for two reasons:
1. The use of `termopen`: If the window containing a terminal buffer opened by termopen is closed then the process is killed, killing the debug adapter. dapui allows toggling of the windows while running so this would break that flow.
2. The existing code assumes that a window exists. This may not be the case for dapui as the windows may not have been opened yet but we still want to open a process to receive output. 

> If the users don't want the window at all they can usually set the debug adapter to run in console mode instead of in integratedTerminal mode.

They do want the window, they just want to be able control it as part of dapui.

> I'm a bit reluctant to allow customizing the integrated terminal behavior more given how core it is to the execution of the debug adapter. It would make triaging issues more difficult.

Totally understandable, alternatively I could create a PR to refactor the existing code to use `jobstart` and not assume a window exists but I wasn't sure if you'd want to change existing behaviour

mfussenegger:
I'll give this some more thought. I tend towards the jobstart approach. Allowing plugins to customize the window/buffer creation is fine but I want nvim-dap core to be in charge of controlling the terminal/process creation.

rcarriga:
Closing this in favour of #554

PR #558, opened 2022-05-24
556 support dap utils pick process on windows

leblocks:


lua/dap/utils.lua
```diff
@@ -34,25 +34,48 @@ end
 
 --- Show a prompt to select a process pid
 function M.pick_process()
-  local output = vim.fn.system({'ps', 'a'})
+  local is_windows = vim.fn.has('win32')
```
mfussenegger, line 37:
```suggestion
  local is_windows = vim.fn.has('win32') == 1
```

`vim.fn.has(..)` returns 0 or 1 and both is truthy in lua

lua/dap/utils.lua
```diff
@@ -34,25 +34,48 @@ end
...
+  local separator = is_windows and ',' or ' \\+'
+  local command = is_windows and {'tasklist', '/nh', '/fo', 'csv'} or {'ps', 'ah'}
```
mfussenegger, line 39: What was the reason to add `h` to `ps a` ?

mfussenegger, line 39: What was the reason to add `h` to `ps a` ?

leblocks: This option disables header line in a _ps_ output, so this check can be removed:
```lua
    if pid and pid ~= 'PID' then
```

Issue #565, opened 2022-05-31
listeners not called if no callback is given for a request


Abh15h3k:
### Debug adapter definition and debug configuration

dap.listeners.after.disconnect.user = function(session, error, response, payload) print("ON DISCONNECT") end

### Steps to Reproduce

start debugger

terminate()

### Expected Result

ON DISCONNECT in `messages`

### Actual Result

nothing

Abh15h3k:
the listeners don't get executed if no callback is set for a given request.

mfussenegger:
Thanks, I merged your PR https://github.com/mfussenegger/nvim-dap/pull/566

PR #566, opened 2022-05-31
set empty callback if no callback given

Abh15h3k:
change for #565


lua/dap/session.lua
```diff
@@ -762,7 +762,7 @@ function Session:handle_body(body)
     self.message_callbacks[decoded.request_seq] = nil
     if not callback then
       log.warn('No callback for ', decoded)
-      return
+      callback = function(...) end --luacheck: no unused args
```
mfussenegger, line 765:
```suggestion
      callback = function() end
```

PR #571, opened 2022-06-01
Fix the precedence operation in session.lua.

asmodeus812:
Adds fix for precedence operation that was not taken into account. On master at the moment the concat takes precedence over the OR operator which was causing a runtime error when the config.name was null, due to trying to concat a nil value.


mfussenegger:
Thanks. I keep making that mistake :/

PR #574, opened 2022-06-03
Fix corner case when both threadId and allThreadsStopped are set

JafarAbdi:
Both `lldb-vscode` and `vscode-cpptools` set both `threadId` and `allThreadsStopped` to `true` after exception/segfault, this was causing the program to continue when trying to expand (in repl's `.threads`) any thread apart from `threadId`

This PR fixes it by checking for `allThreadsStopped` first and then `threadId`

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -526,10 +526,7 @@ function Session:event_stopped(stopped)
     self.stopped_thread_id = stopped.threadId
   end
 
-  if stopped.threadId then
-    progress.report('Thread stopped: ' .. stopped.threadId)
-    self.threads[stopped.threadId].stopped = true
-  elseif stopped.allThreadsStopped then
+  if stopped.allThreadsStopped then
     progress.report('All threads stopped')
     utils.notify(
       'All threads stopped. ' .. stopped.reason and 'Reason: ' .. stopped.reason or '',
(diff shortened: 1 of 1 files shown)


JafarAbdi:
The last commit is for a bug I just found, pthread sometimes return very wired paths (e.g. `/build/glibc-eX1tMB/glibc-2.31/nptl/pthread_create.c`)

lua/dap/session.lua
```diff
@@ -413,7 +413,7 @@ local function jump_to_frame(cur_session, frame, preserve_focus_hint)
     return
   end
   if not source.sourceReference or source.sourceReference == 0 then
-    if not source.path then
+    if not source.path or vim.fn.filereadable(source.path) == 0 then
```
mfussenegger, line 416: This would break cases where the path is a custom uri that can be opened using a BufReadCmd

JafarAbdi: Thanks for the review! 

Do you prefer reverting it? I'm not sure how to handle that.

mfussenegger: Yep please drop the second commit, that will need a different solution.  The first commit looks good 👍

JafarAbdi: Done, thanks!

PR #594, opened 2022-06-29
fix(vscode): insert configurations under name

hown3d:
When running load_launchjs multiple times in 1 session, the configurations are added instead of overwritten.

This PR implements the overwriting current dap configurations by name.

Signed-off-by: Höhl, Lukas <lukas.hoehl@accso.de>


mfussenegger:
This was kinda by design - it's also documented that way. I don't have a strong opinion on the behavior and changing it is fine for me. But it isn't a fix and would require a documentation update.

I'll in any case leave this open for a while so users have a chance to comment.

mfussenegger (approved):

mfussenegger:
Nobody complained so far so I merged this.

PR #598, opened 2022-07-01
tests: add docker container for testing

hown3d:
Tests are currently a bit weird to run locally, because you need to have plenary installed via git and not by e.g. packer.

This PR adds a dockerfile and a script to execute the tests inside an isolated container with nvim, plenary and python installed. 

Signed-off-by: Höhl, Lukas <lukas.hoehl@accso.de>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM alpine:edge
+
+RUN apk add git neovim python3 --update && \
+  # create symlink for python, because lua only calls python not python3
+  ln -s $(which python3) /usr/bin/python 
+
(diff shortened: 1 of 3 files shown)


mfussenegger:
I'd rather not add docker into the mix. Running third party services via docker can be fine and convenient,  running the tests itself via docker containers adds more indirection that tends to make things more complicated instead of less complicated. 

It's not required to fetch plenary via git.  The only requirement is that nvim-dap and plenary are siblings in the fs hierarchy.  See https://github.com/nvim-lua/nvim-lua-plugin-template#run-tests

hown3d:
Alright, thanks for the information!

PR #602, opened 2022-07-09
feat: jump to file instead of source if source matches file content

jonatan-branting:
This PR makes it so that if we stop, and `sourceReference = 1`, we will still jump to the actual file on disk instead of to an unnamed temporary buffer if the file contents match.

This avoids two problems which are present with the current approach:
1. The constant creation of unnamed buffers which have to be manually closed.
2. The loss of buffer-local information such as diagnostics and test progress (e.g. in the case of integration with neotest)

It also contributes to a more seamless experience over all, in my opinion.

(There might be a better way to decide whether or we should jump to the local file or the DAP-supplied source, than to compare the buffer contents. I'm all ears for suggestions!)

If this is something we want to add, is this something we should perhaps hide behind an option?
Another approach would then be to add an option to always goto the file on disk regardless of content differences, if the user so wanted.

What do you think?

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -412,24 +412,35 @@ local function jump_to_frame(cur_session, frame, preserve_focus_hint)
   if preserve_focus_hint or frame.line < 0 then
     return
   end
+
+  local file_bufnr
+  if source.path:match('^([a-z]+)://.*') then
+    file_bufnr = vim.uri_to_bufnr(source.path)
+  else
+    file_bufnr = vim.uri_to_bufnr(vim.uri_from_fname(source.path))
+  end
+
   if not source.sourceReference or source.sourceReference == 0 then
     if not source.path then
       utils.notify('Source path not available, cannot jump to frame', vim.log.levels.INFO)
       return
     end
-    local scheme = source.path:match('^([a-z]+)://.*')
-    local bufnr
-    if scheme then
-      bufnr = vim.uri_to_bufnr(source.path)
-    else
-      bufnr = vim.uri_to_bufnr(vim.uri_from_fname(source.path))
-    end
-    vim.fn.bufload(bufnr)
-    jump_to_location(bufnr, frame.line, frame.column)
+
+    vim.fn.bufload(file_bufnr)
+    jump_to_location(file_bufnr, frame.line, frame.column)
   else
-    cur_session:source(source, function(err, buf)
+    cur_session:source(source, function(err, source_bufnr)
       assert(not err, vim.inspect(err))
-      jump_to_location(buf, frame.line, frame.column)
+
+      local file_content = utils.get_buffer_content_as_string(file_bufnr)
+      local source_content = utils.get_buffer_content_as_string(source_bufnr)
+
+      -- Jump to actual file on disk if source matches file content
+      if file_content == source_content then
+        jump_to_location(file_bufnr, frame.line, frame.column)
+      else
+        jump_to_location(source_bufnr, frame.line, frame.column)
+      end
     end)
   end
 end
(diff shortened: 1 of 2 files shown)


mfussenegger (commented):
The specification is pretty explicit about *not* doing this:

From https://microsoft.github.io/debug-adapter-protocol/specification#Types_Source:

> If sourceReference > 0 the contents of the source must be retrieved through
> the SourceRequest (even if a path is specified).


If the file is available locally, why does the debug adapter not set `sourceReference = 0`  or `nil`?

jonatan-branting:
While I agree that this probably should be up to the server, it does lead to inconsistent behaviour.
Both https://github.com/ruby/debug and the debugger for `dart` does seem to always return `sourceReference = 1` if the source is available in the workspace. There are probably more examples of this.

I can certainly understand if you don't want something like this in `nvim-dap`, but I also see the benefits of allowing this behaviour to be more controllable through the client, as the user is otherwise very much without options if the server acts in a certain (undesired) way.

jonatan-branting:
> If the file is available locally, why does the debug adapter not set `sourceReference = 0` or `nil`?

Great question, is there a specification regarding *how* the server should handle this (i.e. when `sourceReference` should be set). I can only find references to how the client should handle this value.

mfussenegger:
I think part of the reason that the client shouldn't open the file via path if `sourceReference > 0` is to allow debugging files on different machines. If using TCP the debug adapter could run on a server, or even if the debug adapter runs on the same machine, the debugger itself could launch on a remote machine. An example where this gets problematic is if you're stepping through a standard libary routine and you happen to have the same file locally on disk, except a different version.

I think the solution here is to add options to the debug adapters to tell them that everything is running locally.

jonatan-branting:
Yep, I agree. Thanks for the prompt and detailed responses, learned a lot about DAP today :) 

I also got informed on the `ruby/debug` repository that this is indeed possible as well (by passing `localfs = true`).

DanTup:
@jonatan-branting

> the debugger for dart does seem to always return sourceReference = 1 if the source is available in the workspace

👋 I'm the author of the Dart-Code debug adapter and the new adapters in the Dart/Flutter SDK. I'm curious if you have an example of this happening I could look at (and whether you see this using the Dart-Code debug adapters or the newer SDK adapters)? The intention is always to return either a source path or a sourceReference, so if there's a path that's confusing things by doing both I'd like to fix it.

(FWIW in the old adapters we were unable to locate SDK sources locally and should have alway used `sourceReferences` for SDK sources.. but in the new adapters we can better translate SDK libraries to their local paths and instead prefer to use them, so that if you navigated to SDK sources using Go-to-Definition, debug stepping goes into those same local files and not in some magic downloaded sources that clients may not know are the same).

jonatan-branting:
> @jonatan-branting
> 
> 
> 
> > the debugger for dart does seem to always return sourceReference = 1 if the source is available in the workspace
> 
> 
> 
> 👋 I'm the author of the Dart-Code debug adapter and the new adapters in the Dart/Flutter SDK. I'm curious if you have an example of this happening I could look at (and whether you see this using the Dart-Code debug adapters or the newer SDK adapters)? The intention is always to return either a source path or a sourceReference, so if there's a path that's confusing things by doing both I'd like to fix it.
> 
> 
> 
> (FWIW in the old adapters we were unable to locate SDK sources locally and should have alway used `sourceReferences` for SDK sources.. but in the new adapters we can better translate SDK libraries to their local paths and instead prefer to use them, so that if you navigated to SDK sources using Go-to-Definition, debug stepping goes into those same local files and not in some magic downloaded sources that clients may not know are the same).

Hey!

I never personally used the dart server, it was rather something I stumbled into while researching this issue. Sadly, I don't have any links for this.

Might've been something relating to the old adapter, then 🙂

Issue #640, opened 2022-08-10
Make the repl window automatically scroll to the last line


mweisshaupt1988:
### Problem Statement

To be honest I don't know if this is possible in nvim but it would be nice if the repl window would auto scroll to the last line similar to how IDEs do it with the output window.

This pseudo code for example would produce a lot of manual scrolling to follow the progress and would benefit from an autoscroll feature:
for (i=0; i<1000;i++) { println("Currentnumber: $i") }

FYI: I already opened a [feature request](https://github.com/rcarriga/nvim-dap-ui/issues/126) in nvim-dap-ui but this is probably better fitted here.

### Ideas or possible solutions

The only thing I can think of is that it is possible to type "G" to go to the last line in normal mode. Maby something similar is possible to do in an event handler without having focus on the repl window.

mfussenegger:
I had a go at emulating autoscroll similar to a terminal with https://github.com/mfussenegger/nvim-dap/pull/654

But I'm not sure if it's actually a good idea.
I'm wondering if it wouldn't make more sense to instead route the output to a differnt terminal - e.g. the integrated terminal window. That would have the advantage that terminal escape codes would be handled as well.

If you've a use case where the application is constantly writing output you otherwise can't really use the REPL.

Raagh:
The use case I would say is when you are debugging tests. The output of the terminal gets super big and I have to keep following it with my cursor, it is quite annoying

tuxflo:
I think using a terminal buffer would be awesome. Maybe something like ToggleTerm or Neoterm could be used, because then it would be even easier to show/hide the output and configure the size of the buffer easily.

Alphaegen:
@mfussenegger How to actually use this? It doesn't automatically work for me.

drskoolie:
@Alphaegen 

The code that @mfussenegger add allows you to scroll to the bottom of the REPL buffer when you move to it (you can test it out for yourself). 

What we need is an autocommand that triggers when a buffer is updated (even if it's from an external event). I don't know how to do that.

PR #654, opened 2022-08-19
Add autoscroll for the REPL

mfussenegger:
Closes https://github.com/mfussenegger/nvim-dap/issues/640

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -9,6 +9,8 @@ local history = {
   idx = 1
 }
 
+local autoscroll = vim.fn.has('nvim-0.7') == 1
+
 local function get_session()
   return require('dap').session()
 end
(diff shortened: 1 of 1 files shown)


lua/dap/repl.lua
```diff
@@ -9,6 +9,8 @@ local history = {
   idx = 1
 }
 
+local autoscroll = vim.fn.has('nvim-0.7') == 1
```
bselwe, line 12: Hey @mfussenegger, could this support nvim 0.7 and above? So it would support nvim 0.8 as well?

mfussenegger: `vim.fn.has('nvim-0.7') == 1` is true on nvim 0.8 or later as well

patroldo:
Hi, @mfussenegger 

Could you please tell when this commit will be merged? :)
Thanks in advance :)

PR #657, opened 2022-08-20
fix(sign): unplace debugPC sign on dap.close

Kasama:
The `DapStopped` sign is managed by the plugin and placed/unplaced on every `step` or `_jump`.

If the debugged application terminates by itself, it is properly cleaned up.

but if the user calls `require('dap').close()` or `require('dap').terminate()`, the sign is kept in place.

This PR adds a `vim.fn.sign_unplace` call to the `close` session function (which is called both for `close()` and for `terminate()` to properly clean up that sign.

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -987,6 +987,7 @@ function Session.connect(_, adapter, opts, on_connect)
       return
     end
     closed = true
+    vim.fn.sign_unplace(ns_pos)
     client:shutdown()
     client:close()
     session.threads = {}


mfussenegger (commented):
This was only recently intentionally moved to `set_session`, which should be called when the session gets cleared.

See https://github.com/mfussenegger/nvim-dap/commit/674ca42694254ea282f6cb44357dea093d2b29a5 for more context.
Some further background on that is that I intend to add multi-session support at some point in the future, and then doing this within the session.client.close probably wouldn't work anymore.

I also can't reproduce this - at least not with `terminate`. For me it clears the sign.

Kasama:
Hmm I see, the previous change to support multi-session in the future makes sense to me.

Indeed `terminate` works because it's calling [set_session on an event](https://github.com/mfussenegger/nvim-dap/blob/ad8b0de205a077b66cb301531bdc31c8fc7551b6/lua/dap/session.lua#L608). But just calling `require('dap').close()` still leaves the sign behind.

Maybe `close()` should also set the sesion to `nil`? or even call `session:disconnect`, which will check if the user hasn't [started a new session](https://github.com/mfussenegger/nvim-dap/blob/ad8b0de205a077b66cb301531bdc31c8fc7551b6/lua/dap/session.lua#L1409) since closing the previous one

mfussenegger:
> Maybe close() should also set the session to nil? 

That sounds good

It shouldn't call disconnect, see the docs:

>         This does NOT terminate the debug adapter or debugee.
>       You usually want to use either |dap.terminate()| or |dap.disconnect()|
>       instead.

Kasama:
I see. It was probably my bad then and I wanted to use `terminate()` all along instead of `close()`.

Nevertheless, I've updated the MR to `set_session(nil)` on close.

Thanks for your time :smile:

Kasama:
I've also added a test to ensure that happens. Not sure if it's in the proper place, though.

Let me know if it should be somewhere else.

mfussenegger:
I think there was some misunderstanding. I thought that `Maybe close() should also set the session to nil?` was referring to `dap.close`, not to the session.client.close.

Having `session.client.close` call `dap.set_session(nil)` has the aforementioned multi-session problem.

Basically have this instead:

```diff
diff --git a/lua/dap.lua b/lua/dap.lua
index 073a15d..8562858 100644
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -444,11 +444,11 @@ end


 function M.close()
   if session then
     session:close()
-    session = nil
+    M.set_session(nil)
   end
 end


 function M.up()
 ```

Kasama:
:facepalm: indeed it makes total sense, sorry about that.

Should I squash the commits into a single one?

mfussenegger:
> Should I squash the commits into a single one?

No need, I squash merged.

Thanks for the PR

Issue #650, opened 2022-08-17
load_launchjs should use json5 for parsing


gutyina70:
### Problem Statement

Currently a valid launch.json for vscode may not be valid for nvim-dap. For example load_launchjs throws errors when it encounters trailing commas or multiline comments.

### Ideas or possible solutions

Use a library like https://github.com/Joakker/lua-json5 for parsing launch.json instead of neovim's built in json_decode. Usually launch.json files are not too big, so performance shouldn't be an issue. This can also be an optional dependency and keep built in json_decode as a fallback for parsing.

mfussenegger:
I'm not sure I want to pull in lua-json5 as (optional) dependency. It hasn't had a release and the author hasn't been very active in the last couple months, so I'm not sure how solid it is. Even if it is "only" json.

But I could imagine exposing the decode function so that people could override it.

E.g.

```lua
require('dap.ext.vscode').json_decode = json5.parse
```

mwcz:
Exposing the decode function would be really nice.  I work on projects alongside mostly vscode users and it would feel nice to not put restrictions on their launch.json formatting and commenting.

mwcz:
Thank you!

PR #661, opened 2022-08-27
useability: load project local configurations if available.

ldelossa:
Hey, 

At my day job we are configuring remote debugging for our code base. 

Currently, its very easy for us to just add a `launcher.json` file to the repository and everything "just work" when VSCode users open their editors the the source directory. 

Ideally, we'd like this to happen when using nvim-dap as well. 

I understand you can load a launcher.json but there are a few reasons I'm still suggestion loading a project-local nvim-dap.lua configuration. 

1) Loading a launcher.json still involves users creating and sourcing in a .lua configuration file, which some may not even know how to do (with runtime path and import rules etc...). 
2) I simply could not get the same `launcher.json` file we are checking, which works in vscode, to work with nvim-dap.
3) Neovim has an opinionated json decoder and I fear constatly removing trailing "," from config files may become an annoyance (this is not really a huge concern tho). 

This PR will search two directories for a nvim-dap config, "./.nvim-dap.lua" and "./.nvim-dap/nvim-dap.lua" and load them in that preference right before continuing on with the normal flow of the "DapContinue" command. 

The result is pretty nice, we can check in this file, and new neovim can run some `make` targets we have which launch our app in a remote debugger setting, and "DapContinue" just finds the config and connects to the remotes. 

Let me know if you'd be interesting in entertaining this idea.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -245,8 +245,20 @@ local function maybe_enrich_config_and_run(adapter, configuration, opts)
   end
 end
 
+local function load_project_config_if_exists()
+    for _, p in ipairs({"./.nvim-dap.lua", "./.nvim-dap/nvim-dap.lua"}) do
+        local f = io.open(p)
+        if f ~= nil then
+            f:close()
+            utils.notify(string.format("Loading project configuration: %s", p), vim.log.levels.INFO)
+            vim.cmd(":luafile " .. p)
+            return
+        end
+    end
+end
 
 local function select_config_and_run()
+  load_project_config_if_exists()
   local filetype = api.nvim_buf_get_option(0, 'filetype')
   local configurations = M.configurations[filetype] or {}
   assert(


mfussenegger (commented):
Thanks for the PR

I'm a bit mixed on this.

One of the reason I went with `launch.json` support initially was to avoid that projects would have to maintain more editor specific debug configurations. Especially for open source projects with many different contributors it would be kinda annoying to have a configuration for each editor anyone might use.
I felt that removing trailing commas from `launch.json` would be the lower barrier. Especially considering that the JSON standard does not permit them, and tools like the json-language-server report errors about them too.
There would be JSON5 (See also https://github.com/mfussenegger/nvim-dap/issues/650) but for some reason Microsoft choose not to use a `json5` file extension.


That out of the way - I did think of adding some kind of configuration discovery mechanism, but that would probably be more abstract and something that nvim-dap extension authors could also hook into.

E.g. nvim-jdtls could use this to discover main-classes in the project
Other extensions could use it to generate configurations for the test cases and so on.


I'd like to leave this open for a while to see what others think.


---

> Loading a launcher.json still involves users creating and sourcing in a .lua configuration file, which some may not even know how to do (with runtime path and import rules etc...).

Not sure I understand the problem here. If they already configured the adapters, which they need to do, adding the call to load the `launch.json` file shouldn't be much more difficult?


> I simply could not get the same launcher.json file we are checking, which works in vscode, to work with nvim-dap.

Can you elaborate on what exactly wasn't working?
Given that the `launch.json` support wouldn't go away if this is merged, it would still make sense to improve it if it's broken for some reason.

ldelossa:
> Not sure I understand the problem here. If they already configured the adapters, which they need to do, adding the call to load the launch.json file shouldn't be much more difficult?

In our repo, users wouldnt configure the adapters either. The entire process is automated. In context: you can run "make kind" and our tooling launches a kind kubernetes cluster, and then launches our app in a container with Delve already configured in headless mode. This PR makes it so users of nvim dont need to do anything other then install "nvim-dap" plugin, and open Nvim at the root of our source code's repo. 

> Can you elaborate on what exactly wasn't working?
Given that the launch.json support wouldn't go away if this is merged, it would still make sense to improve it if it's broken for some reason.

This im not 100% sure, but i can debug it a bit more. It loaded the file fine, but would not establish the connection. Seemed like it may be a silent error. The details are here: https://github.com/cilium/cilium/pull/21108 which has the "launch.json" along with the corresponding nvim-dap config in a comment: https://github.com/cilium/cilium/pull/21108#issuecomment-1229069196 

Ill try to decipher why it did not work. TBH, I think the lua config to configure nvim-dap is better then trying to reuse launch.js, mostly because if im an neovim user im more inclined to read nvim-dap's README and use the help functionality, but thats subjective. 

> That out of the way - I did think of adding some kind of configuration discovery mechanism, but that would probably be more abstract and something that nvim-dap extension authors could also hook into.

I made this pr in a more or less "quick and dirty" way, saw a quick way to accomplish the goal and went for it. I did have a feeling that the author may want something a bit more sophisticated. If you choose to move forward with the idea, im open to modifying with PR with something that fits your plans for a feature like this. 


> I'd like to leave this open for a while to see what others think

SGTM, looking forward to any responses :)

ldelossa:
One thing in a bit confused about. 

Since I've only used nvim-dap for the purpose of that above PR. 

Since "configurations" are tied to file types, if I were to add a debit configuration for a completely unrelated  Go project, the config will show up in both projects? 

There is no concept of scoping a config to a project at this point, correct? 

Just from a users standpoint, id enjoy if when I call "continue" I only see configs that make sense for the project im opened too.

mfussenegger:
> There is no concept of scoping a config to a project at this point, correct?

Yes, this is what the discovery mechanism I mentioned would help resolve.

ldelossa:
Closing this, as we'll just wrap this functionality in an additional plugin: https://github.com/ldelossa/nvim-dap-projects 
If you see any issues with this approach would appreciate a comment in that PR.

Issue #76, opened 2020-12-04
Make signs configurable


OkanEsen:
Hi! It'd be awesome if we could change the sign symbol and highlight defined here: 

https://github.com/mfussenegger/nvim-dap/blob/master/lua/dap.lua#L103-L105
``` lua
vim.fn.sign_define('DapBreakpoint', {text='B', texthl='', linehl='', numhl=''})
vim.fn.sign_define('DapLogPoint', {text='L', texthl='', linehl='', numhl=''})
vim.fn.sign_define('DapStopped', {text='→', texthl='', linehl='debugPC', numhl=''})
```

I hope I can send you a PR over the weekend but I can't promise anything right now.

Much appreciated, thank you for your time!

mfussenegger:
You could already override the sign definition. You just have to make sure that you do it *after* an initial `require('dap')`.

In your `init.vim` you could do a:


```vimL
lua <<
-- ensure module is loaded
require('dap')

-- override sign definition for fancy breakpoints
vim.fn.sign_define('DapBreakpoint', {text='🛑', texthl='', linehl='', numhl=''})
EOF
```

OkanEsen:
Oh man, I'm stupid, thank you very much. :+1:

PR #664, opened 2022-08-29
user config aware sign define

koopa1338:
if a sign is already set in some user config we do not overwrite the value and skip it. This enables sign configuration by the user regardless where the user requires the dap plugin. This is kind related to the closed issue #76


mfussenegger (commented):
The docs currently mention the order sensitivity and need an update, otherwise this looks good to me. 


Thanks for the PR

koopa1338:
Pushed an update to the docs, would appreciate feedback about wording etc.

mfussenegger:
Looks good to me. Thanks again.

PR #689, opened 2022-09-20
fix(repl): use utils.fmt_error to display errors

williamboman:
Ran into issues with `M.append(err.message)` due to `err.message` being nil (while using https://github.com/microsoft/vscode-js-debug). The server does provide a Message in `body.error` however.


tests/utils_spec.lua
```diff
@@ -40,3 +40,20 @@ describe('utils.non_empty', function()
...
+
+    assert.are.equal('Hello, -!', require('dap.utils').fmt_error('{greeting}, {unbound}!', {
+      greeting = 'Hello',
+    }))
+  end)
+
+  it('interpolates message objects without variables', function ()
+    assert.are.equal('Hello, John!', require('dap.utils').fmt_error('Hello, John!'))
+  end)
+end)
```
williamboman, line 59: I added my own implementation for this first and didn't see one already existed, figured I'd keep the tests

mfussenegger:
Thanks

PR #696, opened 2022-09-25
Breakpoints quickfix list smart replace.

ANtlord:
A list of breakpoints replaces the current quickfix list only if the current quickfix list is a list of breakpoints.  A quickfix list of breakpoints is recognized by the title.  If the title is the same as the title nvim-dap assigns then the list is counted as a quickfix list of breakpoints.  If the current quickfix list consists of something else like results from Telescope or make then the list is saved in the history and nvim-dap creates a new one.


mfussenegger:
Thanks

PR #719, opened 2022-10-12
Add support for env variables reference.

farhanmustar:
Hi, I added the env variables support similar to [vscode](https://code.visualstudio.com/docs/editor/variables-reference#_environment-variables).

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -270,7 +270,7 @@ Things to note:
   - `${workspaceFolder}`: The current working directory of Neovim
   - `${workspaceFolderBasename}`: The name of the folder opened in Neovim
   - `${command:pickProcess}`: Open dialog to pick process using |vim.ui.select|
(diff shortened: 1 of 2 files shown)


doc/dap.txt
```diff
@@ -270,6 +270,7 @@ Things to note:
   - `${workspaceFolder}`: The current working directory of Neovim
   - `${workspaceFolderBasename}`: The name of the folder opened in Neovim
   - `${command:pickProcess}`: Open dialog to pick process using |vim.ui.select|
+  - `${env:HOME}`: The environment variable value similar to |getenv()|
```
mfussenegger, line 273:
```suggestion
  - `${env:Name}`: Environment variable named `Name`, for example: `${env:HOME}`.
```

Otherwise this may give the impression that it is limited to `HOME`.

mfussenegger:
Thanks

PR #724, opened 2022-10-18
Fix incorrect cursor position when opening frame from dapui

jedrzejboczar:
I noticed that sometimes when opening (jumping to) a stack frame in nvim-dap-ui the cursor position is incorrect. Using `nvim_win_call` instead of changing windows fixed the issue.


mfussenegger:
Thanks, merged

PR #744, opened 2022-11-02
Fix REPL throwing an error when response result is empty

wookayin:
The error is when the expression (or statement) in the REPL does not
have any return value, e.g., in a python adapter:

dap> print(2)      # evaluates to None

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -157,8 +157,10 @@ local function evaluate_handler(err, resp)
       tree.render(layer, resp)
     end
   else
-    local lines = vim.split(resp.result, '\n', { trimempty = true })
-    layer.render(lines)
+    if resp.result then
+      local lines = vim.split(resp.result, '\n', { trimempty = true })
+      layer.render(lines)
+    end
   end
 end
 


mfussenegger (commented):
The `result` must be present in the response:

https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Evaluate

```
interface EvaluateResponse extends Response {
  body: {
    /**
     * The result of the evaluate request.
     */
    result: string;
```

This is a bug in debugpy, that was already fixed, but afaik not released yet. See https://github.com/microsoft/debugpy/issues/985

wookayin:
I see. Thanks for the pointer. You would not prefer to be overly protective against buggy adapters? :)

mfussenegger:
> I see. Thanks for the pointer. You would not prefer to be overly protective against buggy adapters? :)

I'd rather not add workarounds. It tends to mask bugs so they don't get fixed or it makes it harder to figure out the problem. E.g. in debugpy using `a = None` and then evaluating `a` was also missing a response (which should return `None`, not nothing)

What would be an acceptable change is to add an assertion that the result is present with a clearer error message that it's the debug adapters fault

wookayin:
Makes a lot of sense. I'll close this then, and will try adding response validation logics in the next time.

Issue #197, opened 2021-06-04
Feature Request: Breakpoint priority


yingzhu146:
Hi there,

Would be great if I could change the breakpoint priority via config - just had to change it manually in the code since some of my other plugins were not well behaved. Exposing changing this via config would be great! :)

 thanks a lot this awesome plugin.

mfussenegger:
In general it sounds good to me. I'll see if there's a chance to make this possible via the `sign_define` override (https://github.com/vim/vim/issues/8334)

cstsunfu:
This is very useful when the diagnostic is conflicted with the breakpoint sign.

entropitor:
@mfussenegger Would you be open to a PR that increases the default sign priority so it's at least above LspError?

mfussenegger:
Closing this here as it's tracked in the vim repo

PR #745, opened 2022-11-02
Make the sign priority of DapBreakpoints higher than DapStopped

wookayin:
Previously, DapBreakpoints (as well as DapBreakpointRejected,
DapBreakpointCondition, etc.) had sign priority of 11
where DapStopped had 12. This means breakpoint signs on the line
where DAP is stopped at would be hidden.

This commit raises the priority of DapBreakpoint signs to 13,
higher than DapStopped so that the breakpoint sign can be visible
on stopped lines. Breakpoints would be more important because
DapStopped would usually have line highlights.

### DIFF
--- a/lua/dap/breakpoints.lua
+++ b/lua/dap/breakpoints.lua
@@ -48,7 +48,7 @@ function M.update(breakpoint)
           ns,
           get_sign_name(bp),
           bp.buf,
-          { lnum = bp.line; priority = 11; }
+          { lnum = bp.line; priority = 13; }
         )
       end
       bp.state.verified = breakpoint.verified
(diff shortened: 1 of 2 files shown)


wookayin:
Related: #197. Would you prefer having priority as a configurable config instead of changing the default? If so, I can change the implementation in such a way.

mfussenegger (commented):
That the stopped indicator has higher priority was done intentionally. Not all color schemes define a highlight for `debugPC` that's distinguishable from the background. 

It's more important to see that it stopped in a line then see if there's a breakpoint - which you'll see again if you move further. 

Regarding making the priority customizable: Yes - but ideally that would happen as part of the sign definition, which would require a upstream change. That would have the advantage that any plugin can use the mechanism and nvim-dap wouldn't require a custom API /configuration for it. See https://github.com/vim/vim/issues/8334

Issue #738, opened 2022-10-27
Node / Webpack Project: paths don't map correctly


jdevries3133:
### Debug adapter definition and debug configuration

See https://github.com/jdevries3133/nvim_config/blob/main/lua/dap_conf.lua

### Debug adapter version

vscode-node2-debug 1.43.0

### Steps to Reproduce

Create a node.js project several folders deep. Create a `launch.json` to debug that server. Create a shared library in an intermediate / adjacent folder. Use tsconfig path aliases to point to the shared library. For example:

```
<root>/.vscode/launch.json
<root>/src/lib/common/foo.js
<root>/src/main/src/index.js
<root>/src/main/package.json
<root>/src/main/webpack.config.js
<root>/src/main/dist/<output>
<root>/src/main/src/<project files>
<root/src/main/tsconfig.json
```

`launch.json` example

```json
{
  "name": "my debug",
  "type": "node",
  "request": "attach",
  "cwd": "${workspaceFolder}/src/main",
  "sourceMaps": true,
  "skipFiles": [
    "<node_internals>/**/*.js"
  ],
  "outFiles": [
    "${workspaceFolder}/src/main/dist/*"
  ]
}
```

`tsconfig.json` example

```json
{
  "compilerOptions": [
    "paths": {
        "lib/*": ["../lib/common/*"]
    }
}
```

Behavior does not change based on presence / absence of `projectRoot`.

I notice that the source map file is outputting some wacky and wild paths. For example, stuff in `lib/common` becomes `<root>/src/main/dist/lib/common/foo.js`. Stuff from the main project becomes something like `<root>/src/main/dist/:webpack:/src/index.js`. I got this information by adding log statements to the dap itself, trying to figure out what was going on. I also see malformed paths like this in the raw source mapping file. 

### Expected Result

I would like to add some heuristics to nvim-dap to try to bridge the gap between these malformed paths and the actual files, so that source mapping works.

### Actual Result

Source mapping does not work, which seems to be caused by deficiencies in node2-debug or webpack or maybe both. If there is another workaround to this problem, I would be overjoyed to hear it! If not, I'd like to try to implement a solution here because this is driving me _crazy_. Clearly, I don't really have all the details, so some help on continuing to debug this would also be awesome.

mfussenegger:
> I would like to add some heuristics to nvim-dap to try to bridge the gap between these malformed paths and the actual files, so that source mapping works.

[...]

PR #751, opened 2022-11-07
Adds mapping of paths between local and remotes

Gordin:
I develop my projects in docker containers, so using nvim-dap wasn't possible, because my local paths are always different from the paths in the containers. I added an optional parameter to dap configurations that allows to have a list of mappings from local to remote paths. It just replaces the start of paths that are sent from the key to the value of the mapping (if there is a match), and vice-verse for the responses from the dap server.

I'm sure there are more/other places where this translation would be needed (is setting breakpoints really the only point where a local path is sent to the server? 🤔), but it works well enough for me. It would be nice if someone who is more familiar with nvim-dap could turn this into a mergeable PR :)

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -206,6 +206,15 @@ This is controlled via a `Configuration`, which has 3 required fields:
     name: string        -- A user-readable name for the configuration.
 <
 
+There is one optional field `remoteDirectoryMapping` that can be used when debugging remotely.
+If set, it will translate between local and remote paths. for example, when your workspace is at
+"/home/abc/work/app", but you are debugging in a container where the code is located at "/app",
+you can add this to the `Configuration`:
+
+>
+    remoteDirectoryMapping = { ["${workspaceFolder}"] = "/app", },
+<
+
 In addition, a `Configuration` accepts an arbitrary number of further options
 which are debug-adapter-specific.
 
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -815,6 +815,7 @@ do
         api.nvim_buf_attach(bufnr, false, { on_detach = remove_breakpoints })
       end
       local path = api.nvim_buf_get_name(bufnr)
+      path = utils.local_to_remote_path(path, self.config['remoteDirectoryMapping'])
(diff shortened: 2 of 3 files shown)


Gordin:
Example of a working dap configuration with this change:
```lua
dap.configurations.typescript = {
  {
    name = 'Debug firebase repl',
    type = 'node2',
    request = 'attach',
    cwd = vim.fn.getcwd(),
    sourceMaps = true,
    port = 9229,
    remoteDirectoryMapping = {
      ["${workspaceFolder}"] = "/app/functions",
    },
    sourceMapPathOverrides = {
      ["${workspaceFolder}/src/**/*.ts"] = "${workspaceFolder}/lib/**/*.js",
      ["*:///app/functions/node_modules/*"] = "${workspaceFolder}/node_modules/*",
      ["*:///./~/*"] = "${workspaceFolder}/node_modules/*",
      ["*://@?:*/?:*/*"] = "${workspaceFolder}/*",
      ["*://?:*/*"] = "${workspaceFolder}/*",
      ["*:///([a-z]):/(.+)"] = "${workspaceFolder}/$1:/$2",
    },
    outFiles = {"${workspaceFolder}/lib/**/*.js"},
  }
}
```
Some/all of the `sourcMapPathOverrides` thing is probably unnecessary, but it works for now, and I'm not touching it...

mfussenegger:
So far path mapping is mostly handled in the debug adapters, not in the clients. Unless other clients start doing that I don't see a reason to add it to nvim-dap.

Afaik that path-mapping doesn't always work is an issue in vscode-node-debug2, and it's successor [vscode-js-debug](https://github.com/microsoft/vscode-js-debug) improved on that. It currently requires [nvim-dap-vscode-js](https://github.com/mxsdev/nvim-dap-vscode-js) to work because it requires off-spec extensions. There are efforts to standardize that (https://github.com/microsoft/debug-adapter-protocol/issues/79, https://github.com/microsoft/vscode-js-debug/issues/1388).

See also https://github.com/mfussenegger/nvim-dap/issues/738#issuecomment-1294717896

PR #770, opened 2022-12-04
Column breakpoint

cosminadrianpopescu:
This PR adds the possibility to also set the column when toggling or setting a breakpoing. 

For example, given this code:

```typescript
const arr = [{name: 'Name 1', id: 1}, {name: 'Name 2', id: 2}];
console.log(arr.map(x => x.id));
```

If I go on the second line and I call `lua require('dap').toggle_breakpoint()`, this would set the breakpoint before logging. This means that the program would stop only once, before the console log, without giving me the possibility to stop and to debug also what happens inside the mapping.

This PR adds the `toggle_breakpoint_column`, with the same parameters as `toggle_breakpoint`. 

So, for example, I could go with the cursor on the `x` from `x.id` (line 2, column 26) and call `lua require('dap').toggle_breakpoint_column()`. This would set the breakpoint inside the `arr.map`. This means that it would stop for each value of the array, and I could debug each value of the array.

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -620,13 +620,13 @@ attach({adapter}, {config})                                       *dap.attach()*
         |dap.continue()| or |dap.run()|
 
 
-set_breakpoint({condition}, {hit_condition}, {log_message})
+set_breakpoint({condition}, {hit_condition}, {log_message}, {column})
                                                           *dap.set_breakpoint()*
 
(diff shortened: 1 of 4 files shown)


mfussenegger:
Thanks for the PR but this would be a breaking change for `toggle_breakpoint` and I've other plans for how to evolve the breakpoint API

Issue #720, opened 2022-10-13
Add hooks for preLaunchTask and postDebugTask


stevearc:
### Problem Statement

The [launch.json file format](https://code.visualstudio.com/docs/editor/debugging#_launchjson-attributes) allows specifying a `preLaunchTask` and a `postDebugTask`, which will run an asynchronous job before and after the debugging session. At the moment, these values are ignored by nvim-dap, which is unfortunate because they may be a required step for making the debug session work.

### Possible Solutions

I have written a task-manager plugin that has support for the VS Code `tasks.json` file format, and currently have [an integration for nvim-dap](https://github.com/stevearc/overseer.nvim/blob/master/doc/third_party.md#dap) that wires these properties up correctly. However, it unfortunately relies on some [pretty gross monkey patching](https://github.com/stevearc/overseer.nvim/blob/844d59f014faa037598aa4d00a08640f58d31390/lua/overseer/init.lua#L226-L253). I'm proposing adding a hook for `dap.run` that will, when present, call a user-defined function and pass it a callback that will resume execution when called. That would be enough for me to get rid of the monkey patching, and it's generic enough that it doesn't tie nvim-dap to overseer, or to the `launch.json` format.

There has been some prior discussion about this (#191), where the suggestion was to put the task running logic into the adapter. I believe that it makes more sense to add this to dap core instead of the adapters because the `preLaunchTask`/`tasks.json` format is universal across adapters.

### Considered Alternatives

We could continue with the monkey patching. It's more brittle, but it does currently work in most cases.

stevearc:
I'm happy to do the work here, btw, I just want to get approval for the approach.

mfussenegger:
I agree that some extension mechanism to handle this use-case would make sense. Do you have a proposal for the API?

Another use-case I can imagine is the `codelldb` adapter where the vscode extension handles a `cargo` property on the client side.
Currently handling this would also work via the adapter `on_config` mechanism, but it doesn't compose well - e.g. a cpp focused dap extension and a rust focused extension may both want to define the `codelldb` adapter, which would then conflict.

So I'm thinking this could either be something where extensions can register handles for specific properties, or it would be a more general pluggable config pre-processing mechanism that's global instead of per adapter.
Not sure how to deal with conflicts if multiple extensions want to handle the same property, or if it is a generic pre-processing how the ordering would work out.

stevearc:
[...]

PR #771, opened 2022-12-05
Support command:pickRemoteProcess

zed0:
## Functionality
The little documented `pickRemoteProcess` command can be used with the `pipeTransport` option to query processes running on a potentially remote host rather than the local machine.

This effectively works by running the command to list processes in the same environment that the debugger will be run in.

See https://code.visualstudio.com/docs/cpp/pipe-transport for details.

## Implementation

In order to pick a remote process it's necessary to know the rest of the `pipeTransport` configuration. This is currently achieved by passing the un-expanded configuration to the `expand_config_variables` function and each of the functions in `var_placeholders_once`. I'm not sure if there's a better way of doing this that would allow placeholders within the `pipeTransport` configuration to be used.

## Reproduction

An example config might look like:
```json
    {
      "name": "C++ Attach in hello_gdb container",
      "type": "cppdbg",
      "request": "attach",
      "program": "/hello.out",
      "processId": "${command:pickRemoteProcess}",
      "pipeTransport": {
          "pipeCwd": "${workspaceRoot}",
          "pipeProgram": "docker",
          "pipeArgs": [
              "exec", 
              "-i", 
              "hello_gdb", 
              "sh",
              "-c"
          ],
          "debuggerPath": "/usr/bin/gdb"
      },
      "sourceFileMap": {
              "/src":"${workspaceRoot}"
      },
      "linux": {
        "MIMode": "gdb"
      },
      "setupCommands": [
        {
          "description": "Enable pretty-printing for gdb",
          "text": "-enable-pretty-printing",
          "ignoreFailures": true
        }
      ]
    }
```
This is adapted from `launch.json` in https://github.com/andyneff/hello-world-gdb which also has a docker container that can be used to test the functionality.

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -274,6 +274,7 @@ Things to note:
   - `${workspaceFolder}`: The current working directory of Neovim
   - `${workspaceFolderBasename}`: The name of the folder opened in Neovim
   - `${command:pickProcess}`: Open dialog to pick process using |vim.ui.select|
+  - `${command:pickRemoteProcess}`: Open dialog to pick a remote process using |vim.ui.select|
   - `${env:Name}`: Environment variable named `Name`, for example: `${env:HOME}`.
 
 ==============================================================================
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -174,9 +174,9 @@ for name in pairs(signs) do
   sign_try_define(name)
 end
 
-local function eval_option(option)
+local function eval_option(option, raw_config)
   if type(option) == 'function' then
-    option = option()
+    option = option(raw_config)
   end
   if type(option) == "thread" then
     assert(coroutine.status(option) == "suspended", "If option is a thread it must be suspended")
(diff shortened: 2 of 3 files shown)


mfussenegger (commented):
Thanks for the PR but I'm afraid I can't merge this. debug adapter specific functionality is out of scope for nvim-dap and the `pipeTransport` property is adapter specific.

You'll instead have to use a function value as property and do this in your configuration or plugin

See also:

- https://github.com/mfussenegger/nvim-dap/discussions/597#discussioncomment-3098086
- https://github.com/mfussenegger/nvim-dap/pull/768
- https://github.com/mfussenegger/nvim-dap/issues/720

zed0:
I had assumed that `pickRemoteProcess` fell into the same category as `pickProcess` which is already implemented.
Thanks for taking a look anyway, I'll have a look at the comments you've provided and update https://github.com/mfussenegger/nvim-dap/wiki/C-C---Rust-(gdb-via--vscode-cpptools) with the results.

zed0:
> I guess nvim-dap could provide a hook to register variable handlers too.
> I'll consider that for https://github.com/mfussenegger/nvim-dap/issues/720

That does sound pretty useful, for all sorts of adapters.

PR #798, opened 2022-12-28
feat(session): pass request seq to callbacks/listeners

rcarriga:
Allows users to distinguish between their own requests and other requests when listening.

The motiviation for this is that nvim-dap maintains some state internally for its functionality (e.g. session.current_frame) which nvim-dap-ui needs to be aware of so users don't have issues with the two plugins showing different information. Some of this state is set after requests such as `stackTrace` for the current frame. nvim-dap-ui needs to listen for this and refresh after it has been processed. Issues arise when the listener causes more stack trace requests, leading to infinite recursive calls because the listener can't distinguish between calls made by nvim-dap and itself.

This is occurring now because of a large refactor of nvim-dap-ui which moves away from maintaining state (which previously used the nvim-dap request result) to making direct calls to the debugging adapter for all data.

For reference here is how I am using this change 
https://github.com/rcarriga/nvim-dap-ui/pull/193/files#diff-d72688c95f8375fabf516ae99d7dc57b58ac3e65f64f5678af05411de83206f9R100-R103
https://github.com/rcarriga/nvim-dap-ui/pull/193/files#diff-d72688c95f8375fabf516ae99d7dc57b58ac3e65f64f5678af05411de83206f9R153-R155

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -1030,12 +1030,13 @@ For example:
 Listeners registered in the `before` table are called *before* the internal `nvim-dap` handlers are called. The listeners registered in the `after` table are called *after* internal `nvim-dap` handlers.
 
 
-For commands (request responses), the listeners are called with four arguments:
+For commands (request responses), the listeners are called with five arguments:
 
 1. The session
 2. An error: A optional table with `message` and `body` properties.
 3. A table with the response. `nil` if an error occurred.
 4. The original payload of the request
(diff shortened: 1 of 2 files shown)


mfussenegger (commented):
Fine for me to make that change. Could you also update the docs under `dap-extensions` ?

rcarriga:
Awesome, done :+1:

PR #819, opened 2023-01-12
fix(switchbuf): use uselast if no buffer found

AckslD:
This is at least my understanding that in core if you have `switchbuf = usetab` and the buffer is not visible `uselast` is used instead.

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -550,10 +550,7 @@ local function jump_to_location(bufnr, line, column, switchbuf, filetype)
       return
     end
   end
-  utils.notify(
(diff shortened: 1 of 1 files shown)


mfussenegger:
I'm not sure about this.  Currently if you want this behaviour you can opt-in using `switchbuf = 'usetab,uselast'` but with this change there wouldn't be a way to opt out

AckslD:
Good point, didn't think about that that would be the same since I use `switchbuf = 'usetab'`, but indeed could instead use `switchbuf = 'usetab,uselast'` :+1:

Issue #833, opened 2023-01-22
:help mappings should provide Lua style default suggestions as well as VimL


feoh:
### Problem Statement

For users new to NeoVim, making the conceptual leap from VimL to Lua can be non trivial, especially since the optimal vim.keymap.set syntax is relatively new to the editor.

It would be great if both styles could be presented.

I've added this for myself so I'll generate a PR from this issue.

### Possible Solutions

Add init.lua style bindings to help mappings for new users.

mfussenegger:
Closed by https://github.com/mfussenegger/nvim-dap/pull/834

PR #834, opened 2023-01-22
#833 - add Lua style example mappings to help

feoh:
As someone very new to NeoVim, I was getting started using [kickstart.nvim](https://github.com/nvim-lua/kickstart.nvim).

As that is a Lua based init script, I couldn't understand how to adapt the VimL nnoremap commands presented in :help mappings into their Lua alternatives. Thankfully in aYoutube video I stumbled into the vim.keymap.set() syntax, but I think presenting both would help new folks out a lot.

Thanks for the amazing plugin!

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -462,8 +462,7 @@ MAPPINGS                                             *dap-mappings*
 
 nvim-dap does not configure any mappings by default.
 
-Some example mappings you could configure:
-
+Some example mappings you could configure, first in traditional VimL, then Lua:
 >vim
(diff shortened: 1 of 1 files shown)


mfussenegger (commented):
Thanks for the PR.

I think given that the rest of the configuration examples are all in Lua, the vimL part could even be removed to only have the Lua version.

feoh:
Good call! I updated the PR accordingly. Love this plugin!

mfussenegger:
Thanks

PR #844, opened 2023-01-29
Add support for running concurrent debug sessions

mfussenegger:
~~Still needs some work~~

New API:

- `dap.sessions()` to return active debug sessions
- `dap.ui.widgets.sessions` to show active debug sessions

Step functions will change the focus automatically if the currently
focused session is not stopped.
This should make common scenarios like debugging client + server where
you step from making requests on the client to receiving request on the
server convenient.

Note that this is unrelated to `startDebugging` support. The PR here is
about concurrent top-level sessions. `startDebugging` support will
introduce hierarchical sessions. (Probably including something like
`children` in the `Session` object)

## Open questions

- ~~If there is an active session, should `run_last` / `run` default to restart (terminate, run new), or should they default to run an additional session?~~ Will restart if config name matches, otherwise start  additional session

https://user-images.githubusercontent.com/38700/215311877-f07788e5-348b-4d6c-aaea-188005cff981.mp4

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -615,7 +615,7 @@ continue()                                                      *dap.continue()*
         application.
 
 
-run({config})                                                        *dap.run()*
+run({config}, {opts})                                                *dap.run()*
         Looks up a debug adapter entry for the given configuration and runs it.
         This is implicitly called by |dap.continue()| if no debug session is
         active.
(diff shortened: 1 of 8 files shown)


mfussenegger:
@rcarriga heads up so you can check nvim-dap-ui with these changes as I'm probably going to merge this soonish.  Overall for most users nothing should really change about their existing workflows. But you may want to create a sessions widget and one important difference is that if users do spawn multiple sessions, there can still be active sessions after receiving a terminated event (so closing the dapui on it would be wrong,  one would have to check via dap.sessions() if it was the last)

rcarriga:
Thanks for tagging! So there's no breaking change here, just as you say it'd be useful for users to be able to choose a session?

mfussenegger:
> just as you say it'd be useful for users to be able to choose a session?

Yes and that in general there could now be more than one active session. That affects listener triggers and so on

PR #852, opened 2023-02-07
Prompt once per input + support label/value opts

giancarlo-misasi:
* Small change to support options with key/value pairs instead of just string
* Changed inputs to collect once per input, instead of once per occurence

### DIFF
--- a/lua/dap/ext/vscode.lua
+++ b/lua/dap/ext/vscode.lua
@@ -12,18 +12,21 @@ local function create_input(type_, input)
       if not vim.endswith(description, ': ') then
         description = description .. ': '
       end
-      return vim.fn.input(description, input.default)
+      return vim.fn.input(description, input.default or '')
     end
   elseif type_ == "pickString" then
     return function()
       local options = assert(input.options, "input of type pickString must have an `options` property")
       local opts = {
-        prompt = input.description
+        prompt = input.description,
+        format_item = function(option)
+          return option["label"] or option
+        end
       }
       local co = coroutine.running()
       vim.ui.select(options, opts, function(option)
         vim.schedule(function()
-          coroutine.resume(co, option or input.default)
+          coroutine.resume(co, option["value"] or option or input.default or '')
         end)
       end)
       return coroutine.yield()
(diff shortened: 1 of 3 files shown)


mfussenegger (commented):
A doc update for the change to support objects and some tests would be nice

lua/dap/ext/vscode.lua
```diff
@@ -60,30 +63,70 @@ local function chain(default, fns)
 end
 
 
-local function apply_input(inputs, value)
+local function add_vscode_predefined(once)
+    -- https://code.visualstudio.com/docs/editor/variables-reference
+    -- https://github.com/EthanJWright/vs-tasks.nvim/blob/main/lua/vstask/Predefined.lua
+    -- TODO: Implement more prefefined, workspaceFolder is the most useful
+    once["${workspaceFolder}"] = vim.fn.getcwd()
```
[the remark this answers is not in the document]
giancarlo-misasi: I wasn't aware this was a feature, that's awesome it already is. I will remove this from the PR once I have an update with documentation and tests.

lua/dap/ext/vscode.lua
```diff
@@ -96,8 +126,9 @@ end
 
 local function apply_inputs(config, inputs)
   local result = {}
+  local once = {}
   for key, value in pairs(config) do
-    result[key] = apply_input(inputs, value)
+    result[key] = apply_input(once, inputs, value)
```
mfussenegger, line 131: Doesn't this freeze the value for the entire neovim lifetime?

mfussenegger (commented):
I suspect the `once` handling would freeze input values for the full neovim lifetime. I'm not sure if that's expected?

I extracted the other two changes into https://github.com/mfussenegger/nvim-dap/pull/874

lua/dap/ext/vscode.lua
giancarlo-misasi: I've been using this implementation for the last few weeks, and it should prompt again the next time you start the debugger, at least that's been my experience. The intention is only to avoid re-entering inputs which are repeated in the json. When I have a chance I can add some tests to validate that behavior.

mfussenegger:
Closing this for now.  Please re-open once you had time to investigate and have an update

PR #855, opened 2023-02-10
Add support for LuaRocks + release workflow

mrcjkb:
nvim-dap is a common dependency across Neovim plugins. Using luarocks may alleviate the need for users to specify their plugins' dependencies in their plugin manager (e.g., vim-plug or packer).
See also [this blog post](https://teto.github.io/posts/2021-09-17-neovim-plugin-luarocks.html) for details.

This PR adds a release workflow that publishes this plugin to [LuaRocks](https://luarocks.org/) whenever a tag is pushed, as well as a rockspec that can be used to release to the `dev` channel or install this plugin from source using the luarocks CLI.

For the release workflow to work, someone with a LuaRocks account will have to add their API key to this repo's secrets.

Note that I have added a shield to the readme that assumes the existence of an `mfussenegger/nvim-dap` LuaRocks module.

### DIFF
--- a/nvim-dap-scm-1.rockspec
+++ b/nvim-dap-scm-1.rockspec
@@ -1,0 +1,40 @@
+local _MODREV, _SPECREV = 'scm', '-1'
+rockspec_format = "3.0"
+package = 'nvim-dap'
+version = _MODREV .. _SPECREV
+
+description = {
+  summary = 'Debug Adapter Protocol client implementation for Neovim.',
+  detailed = [[
+  nvim-dap allows you to:
+
+  * Launch an application to debug
+  * Attach to running applications and debug them
+  * Set breakpoints and step through code
+  * Inspect the state of the application
+  ]],
+  labels = {
+    'neovim',
+    'plugin',
+    'debug-adapter-protocol',
+    'debugger',
+  },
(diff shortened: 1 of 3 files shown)


mfussenegger (commented):
Seems okay to me on a first glance. I'll have a closer look before I cut the next release

nvim-dap-scm-1.rockspec
```diff
@@ -0,0 +1,40 @@
...
+
+source = {
+   url = 'git://github.com/mfussenegger/nvim-dap',
+}
+
+build = {
+   type = 'builtin',
+   copy_directories = {
+     'doc',
+     'plugin',
```
mfussenegger, line 38: Does this have to include the lua folder too, or is that picked up by default?

mrcjkb:
Thanks for supporting this!

Yes, I think that's generally good advice if you depend on a plugin that is more than just a library.


I'm just hopeful that it will one day be easier (and maybe common practice) to leverage the dependency resolution capabilities of luarocks.

(See also [my post on some of the pain points](https://mrcjkb.github.io/posts/2023-01-10-luarocks-tag-release.html) that follows up on @teto's posts).

Issue #898, opened 2023-03-15
Program suddenly exit when call `si` or `n` or other continue command in repl.


Katzeee:
### Debug adapter definition and debug configuration

Installed adapter via `Mason`, I tried `cpptools` and `codelldb`, both have this issue.
Here is my config, should the same as wiki's.
```lua
return {
   adapters = {
     codelldb = {
       type = "server",
       port = "${port}",
       executable = {
          command = codelldb_path,
          args = { "--port", "${port}" },
        },
      },
      cppdbg = {
           id = "cppdbg",
           type = "executable",
          command = install_root_dir .. "/bin/OpenDebugAD7",
       },
  },
  configurations = {
    cpp = {
      name = "Launch File",
      type = "cppdbg",
      -- type = "codelldb",
      request = "launch",
      program = function() return vim.fn.input("Path to executable: ", vim.fn.getcwd() .. "/", "file") end
      cwd = "${workspaceFolder}",
      stopOnEntry = false,
    },
  },
}
```

### Steps to Reproduce

```cpp
#include <iostream>
#include <memory>
int f() {
  static int x = 0;
  x++;
  return 1;
}
int main() {
  int x = 0;
  for (int i = 0; i < 5; i++) {
    x += 1;
  }
  f();
  auto node = std::make_unique<int>(1);
  std::cout << &node << " " << node.get() << std::endl;
}
```
I set the breakpoint at line 5, then `dap.continue`, `dap.repl.open`, call `-exec si`, the program suddenly exit.

### Expected Result

I expect the program stop at this line, repl output something like
```
0x00005555555551c3    6    x++;
```
as it output when I debug it using `gdb` or `lldb`.
I find the output message and a stop event in `dap.log`, but it is wierd that I also got this session debug message in it.
```
[ DEBUG ] 2023-03-15T08:30:02Z+0800 ] .../xac/.local/share/nvim/lazy/nvi
m-dap/lua/dap/session.lua:932 ] 1       {
  body = {
    category = "stdout",
    output = "0x00005555555551c3\t6\t  x++;\n"
  },
  event = "output",
  seq = 43,
  type = "event"
}
[ DEBUG ] 2023-03-15T08:30:02Z+0800 ] .../xac/.local/share/nvim/lazy/nvi
m-dap/lua/dap/session.lua:932 ] 1       {
  body = {
    allThreadsStopped = true,
    column = 1,
    line = 6,
    reason = "step",
    source = {
      checksums = {},
[...]

PR #899, opened 2023-03-15
Fix repl single step command cause exit and no jump_frame

Katzeee:
As #898 says, when I am debuggin a single thread cpp program, call `-exec n` with `cpptools` or `n` with `codelldb` makes program suddenly exit.
I found `auto_continue_if_many_stopped` is set for multi-thread debugging, you don't want to jump to the `stopped.threadId` if we now hold a stopped_thread, but the jumping is reasonable if this `StoppedEvent` occurs at the thread we are now hold.

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -641,7 +641,9 @@ function Session:event_stopped(stopped)
   end
 
   local should_jump = stopped.reason ~= 'pause' or stopped.allThreadsStopped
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks for the report and PR. I wanted to add a test case and change the comment slightly but couldn't push on the PR. So I merged this via https://github.com/mfussenegger/nvim-dap/pull/901 instead

PR #911, opened 2023-03-24
Pass parent session to adapter factory

mfussenegger:
For the `startDebugging` implementation of `vscode-js-debug` the client
needs to connect to the parent adapter instance.

This change enables an adapter configuration like this:

```lua
require("dap").adapters["pwa-node"] = function(on_config, config, parent)
  local target = config["__pendingTargetId"]
  if target and parent then
    local adapter = parent.adapter --[[@as ServerAdapter]]
    on_config({
      type = "server",
      host = "localhost",
      port = adapter.port
    })
  else
    on_config({
      type = "server",
      host = "localhost",
      port = "${port}",
      executable = {
        command = "node",
        args = {"/path/to/js-debug/src/dapDebugServer.js", "${port}"},
      }
    })
  end
end
```

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -143,9 +143,13 @@ Both types support the following additional options:
                                        -- retrieved via a source request.
 
 
-`dap.adapters.<name>` can also be set to a function which takes two arguments.
-This first argument is a callback which must be called with the adapter table.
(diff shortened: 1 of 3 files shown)


mfussenegger:
@mxsdev fyi: The example adapter configuration and this change will probably be relevant for nvim-dap-vscode-js

PR #916, opened 2023-04-02
Allow process filtering for utils.pick_process

Mnib:
Hi,

This PR enables the option to filter the processes returned by `utils.pick_process` by passing these opts:
```
opts = {
  match_pattern = "string"
}
```

I made this PR because I use the pick process feature to attach an adapter to a process running my tests, and this would reduce the number of processes returned and make the picking a bit more friendly.

Please let me know if you have any suggestions on this PR or if some things should be changed


mfussenegger:
Thanks for the PR. I went with a slightly different approach. See https://github.com/mfussenegger/nvim-dap/pull/922

Given that the PR was created from master I couldn't amend the changes here

PR #917, opened 2023-04-02
docs: add installation instructions for lazy.nvim

Haris-AK:
Added instructions in the README for how to use nvim-dap with Lazy.nvim as a package manager.


mfussenegger:
Thanks but I think I'll instead go into the opposite direction and just document the stock neovim plugin installation and have a generic "or with your package manager".

PR #937, opened 2023-04-28
Make terminate behavior on vim exit configurable per adapter

Kareeeeem:
When developing I treat vim instances as disposable, opening and closing them as I go. When I have the debuggee running as a separate process I don't always want it to terminate if vim exits with an open session. 

I have 0 experience with lua or vim plugin development but I tried to make this configurable with a new option on the adapter. I'm not sure how to test the exit handler.

I'm not very happy with the negative phrasing of the option. I initially had it as `terminate_on_exit`, but wasn't sure how to actually default this configuration, and this meant `nil` and `true` were the same when it's eventually checked and I did not like that. So I made the option negative.

this is an example config I use 

```
table.insert(py_configs, {
    type = 'python';
    request = 'attach';
    name = 'Attach Running APP';
    dont_terminate_on_exit = true,
    connect = {
        host = '127.0.0.1',
        port = '${env:DEBUGPY_PORT}',
    },
})
```

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -204,6 +204,12 @@ This is controlled via a `Configuration`, which has 3 required fields:
     name: string        -- A user-readable name for the configuration.
 <
 
+There is one optional field to control whether the debuggee is terminated when
+vim exits:
+
+>
+    dont_terminate_on_exit: bool        -- defaults to false
+<
 In addition, a `Configuration` accepts an arbitrary number of further options
 which are debug-adapter-specific.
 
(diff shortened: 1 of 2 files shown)


mfussenegger:
The payload of configurations is for the debug adapter, not for the client configuration.

This also seems pretty niche, so I'm not sure if this should be built-in. Maybe create an issue first to gauge interest with upvotes.

mfussenegger:
You could try adding a `QuitPre` or `ExitPre` autocmd that calls dap.disconnect(). If the session is already closed by the time `VimLeavePre` fires it can't terminate it anymore.

PR #942, opened 2023-05-07
remove REPL history clear when starting a new session

atouminet:
Currently starting a new session clears the REPL history. I see no reason why this is required, and actually I have found quite common the need to reuse REPL command between sessions.
Other IDEs (e.g. vscode) also keep history between sessions.

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -7,7 +7,7 @@ local history = {
   last = nil,
   entries = {},
   idx = 1,
-  max_size = 5000,
+  max_size = 100,
 }
 
 local autoscroll = vim.fn.has('nvim-0.7') == 1
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -6,7 +6,8 @@ local M = {}
 local history = {
   last = nil,
   entries = {},
-  idx = 1
+  idx = 1,
+  max_size = 5000,
 }
 
 local autoscroll = vim.fn.has('nvim-0.7') == 1
@@ -226,6 +227,9 @@ function execute(text)
     end
   else
     history.last = text
+    if #history.entries == history.max_size then
+      table.remove(history.entries, 1)
+    end
     table.insert(history.entries, text)
     history.idx = #history.entries + 1
   end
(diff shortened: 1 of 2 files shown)


mfussenegger:
This would also stop clearing the output, which I very much want to keep. 

I don't mind not clearing the history, but then it would probably need to become a ringbuffer to avoid the memory usage from growing indefinitely.

atouminet:
I have made changes according to your review. I have added a maximum REPL history size, with a hardcoded max size.
I can make it a configurable option if needed, although I do not think it is very useful to have a much larger history.

lua/dap/repl.lua
```diff
@@ -245,8 +249,7 @@ function execute(text)
     return
   elseif vim.tbl_contains(M.commands.clear, text) then
     if repl.buf and api.nvim_buf_is_loaded(repl.buf) then
-      local layer = ui.layer(repl.buf)
-      layer.render({}, tostring, {}, 0, - 1)
+      M.clear_output()
```
mfussenegger, line 252:
```suggestion
      M.clear()
```

No reason the break the API

lua/dap/repl.lua
```diff
@@ -383,10 +386,7 @@ function M.append(line, lnum, opts)
 end
 
 
-function M.clear()
-  history.last = nil
-  history.entries = {}
-  history.idx = 1
+function M.clear_output()
```
mfussenegger, line 389:
```suggestion
function M.clear()
```

lua/dap/session.lua
```diff
@@ -1623,7 +1623,7 @@ end
 --- Initialize the debug session
 ---@param config Configuration
 function Session:initialize(config)
-  vim.schedule(repl.clear)
+  vim.schedule(repl.clear_output)
```
mfussenegger, line 1626:
```suggestion
  vim.schedule(repl.clear)
```

lua/dap/repl.lua
```diff
@@ -6,7 +6,8 @@ local M = {}
 local history = {
   last = nil,
   entries = {},
-  idx = 1
+  idx = 1,
+  max_size = 5000,
```
mfussenegger, line 10:
```suggestion
  max_size = 100,
```

lua/dap/session.lua
daoye: Is it possible to add an option to control whether to clear the history when start a new session?

PR #952, opened 2023-05-19
Implement `run_launchjs` for running `launch.json` configs interactively

utilyre:
- The reasoning is explained in commit messages.

- Why are comments stripped out in `load_launchjs` even though `json` (unline `jsonc`) does not support comments?


daoye:
That's really good.

PR #969, opened 2023-06-10
fix: Exit code 1 warning at the end of a debug session on Windows

Aeceon:
On Windows, when using a server type debug adapter that is executed (in this case codelldb) I received the following warning after a debug session ended:

> ...codelldb.CMD exited with code 1

Changing "sigterm" to "SIGTERM" in the client close function fixes this.

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -1201,9 +1201,14 @@ local function spawn_server_executable(executable, session)
   stderr:read_start(read_output('stderr', stderr))
   stdout:read_start(read_output('stdout', stdout))
 
+  local is_windows = vim.fn.has("win32")
   session.on_close["dap.server_executable"] = function()
     if not handle:is_closing() then
-      handle:kill("sigterm")
+      if is_windows == 1 then
+        handle:kill("sighup")
+      else
+        handle:kill("sigterm")
+      end
     end
   end
 end
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -1206,7 +1206,7 @@ function Session.connect(_, adapter, opts, on_connect)
       close = function(cb)
         _close(function()
           if not handle:is_closing() then
-            handle:kill("sigterm")
+            handle:kill("SIGTERM")
           end
           if cb then
             cb()
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -1202,11 +1202,16 @@ function Session.connect(_, adapter, opts, on_connect)
     session.adapter = adapter
 
     if handle then
+      local is_windows = vim.fn.has("win32")
       local _close = close
       close = function(cb)
         _close(function()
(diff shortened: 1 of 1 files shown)


Aeceon:
I've jumped the gun here. I should have opened an issue first.

After receiving the exit code 1 warning, I read the Neovim documentation on kill and noticed that sigterm was in uppercase [here](https://neovim.io/doc/user/luvref.html#luv-signal-handle). I made [this change](https://github.com/mfussenegger/nvim-dap/pull/969/files#diff-1c238a5ffb6d36f9412528c8783aacf498b7c4456f7ce2131779e1ea04bb98b2) and the debug adapter correctly exited with code 0. I should have looked into it further because It correctly exits with code 0 on Windows for everything but "sigterm" (lowercase). For instance:

```lua
handle:kill("blah")
```

This sees the debug adapter exit with code 0 as well.

### Steps to reproduce the original issue on Windows (not WSL):

1. The following adapter configuration automatically provided by mason-nvim-dap:
```lua
local dap = require('dap')

dap.adapters.codelldb = {
	type = 'server',
	port = '${port}',
	executable = {
		command = vim.fn.exepath('codelldb'),
		args = { '--port', '${port}' },
	},
}

if vim.fn.has('win32') == 1 then
	dap.adapters.codelldb.executable.detached = false
end

dap.configurations.cpp = {
  {
    name = 'Launch',
    type = 'codelldb',
    request = 'launch',
    program = function()
      return vim.fn.input('Path to executable: ', vim.fn.getcwd() .. '/', 'file')
    end,
    cwd = '${workspaceFolder}',
    stopOnEntry = false,
    args = {},
}

dap.configurations.c = dap.configurations.cpp
dap.configurations.rust = dap.configurations.cpp
```
2. Compile a standard hello world program in C with MSVC ("cl /Zi main.c").
3. Open the hello world source file in Neovim.
4. Set a breakpoint before print.
5. Start a debug session with the compiled program.
6. Hit continue once the breakpoint has been reached.

I then receive the following warning when the debug session ends (generated on [line 1152 of session.lua](https://github.com/mfussenegger/nvim-dap/blob/master/lua/dap/session.lua#L1152)):
>...codelldb.CMD exited with code 1

This doesn't happen if handle:kill is passed anything other than "sigterm" (lowercase).

mfussenegger:
Could you try with `sigint` ?

Changing it to `SIGTERM` prevents it from terminating the process on linux, and it again leaks processes if they don't stop by themselves after a debug session (which vscode-js-debug doesn't).

mfussenegger:
@Aeceon did you have a chance to try with `sigint`?

Aeceon:
Sorry, I forgot I hadn't updated this. `sigint` gave the same "exited with code 1" error. `sighup`, however, works. I changed the code to only use this on Windows and stick to `sigterm` otherwise.

PR #996, opened 2023-07-27
support register variable/command, add unit test

AbaoFromCUG:
# Register command mechanism
Regarding to [Command](https://code.visualstudio.com/api/extension-guides/command#creating-new-commands), vscode has a mechanism named `Command`, which can be used as [Command Variable](https://code.visualstudio.com/docs/editor/variables-reference#_command-variables)
> If the predefined variables from above are not sufficient, you can use any VS Code command as a variable through the ${command:commandID} syntax.
> A command variable is replaced with the (string) result from the command evaluation. The implementation of a command can range from a simple calculation with no UI, to some sophisticated functionality based on the UI features available via VS Code's extension API. If the command returns anything other than a string, then the variable replacement will not complete. Command variables must return a string.

For example, [vscode-cmake-tools](https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/debug-launch.md#gdb) extension support lots of  `command variable`, which can be used in launch.json

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "(gdb) Launch",
            "type": "cppdbg",
            "request": "launch",
            // Resolved by CMake Tools:
            "program": "${command:cmake.launchTargetPath}",
            "cwd": "${workspaceFolder}",
            "environment": [
                {
                    // add the directory where our target was built to the PATHs
                    // it gets resolved by CMake Tools:
                    "name": "PATH",
                    "value": "${env:PATH}:${command:cmake.getLaunchTargetDirectory}"
                }
            ]
        }
    ]
}
```

# What did I do

I add API named `dap.register_command(identifier, handler)` to do something same to [vscode.commands.registerCommand](https://code.visualstudio.com/api/references/vscode-api#commands.registerCommand) .  Which can help extension developer to enhance ability of dap.

 e.g.
  ```lua
require("dap").register_command("cmake.launchTargetPath", function()
     // run cmake build task, then return executable path
     // can be a asynchronous coroutine
     return coroutine.create(function()
           // fake cmake build step
           vim.defer_fn(function()
              coroutine.resume(co, "/path/to/exectutable")
            end, 2000)
    end)
end)
```

# Some Unit Test Added

* normal variable
* env varialbe
* register command
* register asnyc command

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -264,9 +264,9 @@ for name in pairs(signs) do
   sign_try_define(name)
 end
 
-local function eval_option(option)
+local function eval_option(option, ...)
   if type(option) == 'function' then
-    option = option()
+    option = option(...)
   end
   if type(option) == "thread" then
     assert(coroutine.status(option) == "suspended", "If option is a thread it must be suspended")
@@ -274,8 +274,9 @@ local function eval_option(option)
     -- Schedule ensures `coroutine.resume` happens _after_ coroutine.yield
     -- This is necessary in case the option coroutine is synchronous and
     -- gives back control immediately
+    local arg = { ... }
     vim.schedule(function()
-      coroutine.resume(option, co)
+      coroutine.resume(option, unpack(arg))
     end)
     option = coroutine.yield()
   end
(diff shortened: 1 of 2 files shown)


AbaoFromCUG:
Closed with additional inject code via `dap.expand_config_variables`

Issue #737, opened 2022-10-26
Control the output verbosity or how expressions are printed in DAP REPL


wookayin:
### Problem Statement

Can we control how an evaluated expression is printed in the DAP REPL? If the default one is too verbose, showing all the attributes (as a tree) expanded, can we make it compact so that it behaves closer to conventional debuggers --- specifically, when it comes to python, like pudb or ipdb.

(This might be a bit python-specific, but could be related to other languages as well.)

<img width="1244" alt="image" src="https://user-images.githubusercontent.com/1009873/197931620-8e92c246-3e36-41aa-8e32-9dc7246b20ad.png">

For instance, when printing out some python object in the REPL, it shows not only the **`__repr__`** representation but also all the attributes as in the variables/watches widget. As an example, a `pandas.DataFrame` object will print ~100 attributes and lines that would be too much verbose and not that quite informative.

### Possible Solutions

(1) Always print the attribute tree in the REPL window, but make it collapsible so that when collapsed it takes at most one line. In this particular example, it may show a few, concise lines like (which could be collapsible and expandable):
```
▸ df DataFrame = Empty DataFrame
                 Columns: []  
                 Index: []    
```

(2) Add an option (in any ways) to completely disable displaying the attributes; and print the `__repr__(object)` only as ipdb, pudb, etc. does.

<img width="1033" alt="image" src="https://user-images.githubusercontent.com/1009873/197932063-6ea6a9aa-55d3-4540-ba8c-78daacf568f9.png">

### Considered Alternatives

microsoft/debugpy#1078

mfussenegger:
Relates to https://github.com/mfussenegger/nvim-dap/issues/1062

This seems a bit specific to python. Only printing the root would make it significantly worse for other languages. 

Always printing `repr(object)` would require language specific knowledge in nvim-dap. That's out of scope but
a fairly simple solution would be the `.e0` repl command mentioned in the other issue. It would give users control over how many levels to print

Other than that I think it would require some custom formatting hooks. I'm open to suggestions how those could look like

wookayin:
Is it difficult to make the tree collapsible by default or upon configuration? Or any kind of configuration point will be great.

carschandler:
I agree with this completely!

Here are a few examples of how other debugger environments handle this:

[...]

PR #1001, opened 2023-07-29
Add a new option dap.repl.show_tree: whether to show the tree in REPL

wookayin:
A new option `show_tree` can be set by users to configure whether to
show the expandable tree in the DAP REPL window. Concise and simple
representation in the REPL would be useful for many languages,
see #737 for an example (python).

The option can be configured using, for instance:

```
require('dap.repl').show_tree = false
```

TODO: Add documentation in `doc/dap.txt`

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -11,6 +11,7 @@ local history = {
 }
 
 local autoscroll = vim.fn.has('nvim-0.7') == 1
+M.show_tree = true
 
 local function get_session()
   return require('dap').session()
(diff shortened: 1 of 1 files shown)


wookayin:
@mfussenegger Kindly bumping up and bringing this again to your attention.

mfussenegger:
Thanks for the PR. I think having more options to control how the variables are displayed is something I want to add but this is rather limited and I'm not sure about the UI/UX. 

I'll need to give this some more thought but I'm pretty sure it will be different than this,  so I'm closing the PR

wookayin:
Hope this or similar features can be implemented soon -- this is so essential for python debugging. I'm happy to discuss and contribute more.

wookayin:
The reason I chose somewhat limited configuration like this is that nvim-dap doesn't have an universal configuration API, and previously we used to inject user configs into global variables (e.g. https://github.com/mfussenegger/nvim-dap/blob/master/doc/dap.txt#L564-L570) We wouldn't need a UI for configuration.

PR #1024, opened 2023-08-28
Stop terminating "attached" debugees when exiting Vim

NSteinhoff:
Hi,

I mostly use DAP to debug a long-running server process. My workflow has me closing Vim frequently, and whenever I do without remembering to disconnect the debugger, I accidentally terminate the server. Is that intended behavior? Why would DAP want to manage the lifecycle of an "attached" debugee?

Anyways, this hack seems to fix it for me. This was the smallest change I could think of, but it's a breaking change. Perhaps it would be better placed as a per configuration option (`terminate_on_vim_exit = false` or `on_vim_exit = "terminate" | "disconnect"`), which defaults to termination, but allows opting out?

I'm not particularly attached to any of this. Take it or leave it, or modify it as you see fit :)

Thanks for your work on this plugin 👍

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -1038,12 +1038,12 @@ function M._vim_exit_handler()
       s:disconnect({ terminateDebuggee = false })
     else
       terminate(s)
-      vim.wait(500, function()
-        ---@diagnostic disable-next-line: redundant-return-value
-        return session == nil and next(sessions) == nil
-      end)
     end
   end
+  vim.wait(500, function()
+    ---@diagnostic disable-next-line: redundant-return-value
+    return session == nil and next(sessions) == nil
+  end)
   M.repl.close()
 end
 
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -1034,11 +1034,15 @@ end
 
 function M._vim_exit_handler()
   for _, s in pairs(sessions) do
-    terminate(s)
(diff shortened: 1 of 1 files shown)


lua/dap.lua
```diff
@@ -1034,11 +1034,15 @@ end
...
-    end)
+    if s.config.request == "attach" then
+      s:disconnect({ terminateDebuggee = false })
+    else
+      terminate(s)
+      vim.wait(500, function()
+        ---@diagnostic disable-next-line: redundant-return-value
+        return session == nil and next(sessions) == nil
+      end)
+    end
```
mfussenegger, line 1045:
```suggestion
    if s.config.request == "attach" then
      s:disconnect({ terminateDebuggee = false })
    else
      terminate(s)
    end
    vim.wait(500, function()
      ---@diagnostic disable-next-line: redundant-return-value
      return session == nil and next(sessions) == nil
    end)
```

mfussenegger (commented):
I think it should be okay to make this change. 

At least I currently can't think of a scenario where this would leak a process also created by nvim-dap, but I may be missing something obvious.

In any case, maybe worth a try. If somebody complains it can still be reverted.
(And in that case, you could try adding a `QuitPre` autocommand that disconnects. If by the time `ExitPre` fires the session is already gone, nvim-dap won't terminate it)

PR #1030, opened 2023-09-01
Support source maps between editor and debugger file paths

gpanders:
When nvim-dap is running on a separate machine from the debugger then Nvim and the debugger will use different paths for source files. For example, if Nvim is running on a machine where the source tree is located at `/a/project/foo` and the debugger is running on a separate machine where the source tree is located at `/b/project/foo`, when nvim-dap tries to set a breakpoint it will use `/a/project/foo/` in the breakpoint location. When the debugger receives this request it will not set the breakpoint correctly because the debugger does not have any files at `/a/project/foo`.

The `source_map` option instructs nvim-dap to convert paths as seen by the editor and nvim-dap itself to the paths as the debugger sees them. The opposite mapping takes place when nvim-dap receives stack frame or source file information from the debugger.

Note that this is different than the "source map" option provided by many debuggers themselves (such as codelldb's "sourceMap" option or GDB's "substitute-path" setting). Those options convert _compiled_ source paths (i.e. the path of the source file when the binary was compiled) into the debugger's path namespace (`/b/project/foo` in our example).

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -155,17 +155,23 @@ All types support the following additional options:
 
 >
     options?: {
-      initialize_timeout_sec?: number  -- How many seconds the client waits for a
-                                       -- response on a initialize request before
-                                       -- emitting a warning. Defaults to 4
+      initialize_timeout_sec?: number    -- How many seconds the client waits for a
+                                         -- response on a initialize request before
+                                         -- emitting a warning. Defaults to 4
 
-      disconnect_timeout_sec?: number  -- How many seconds the client waits for
-                                       -- a disconnect response from the debug
-                                       -- adapter before emitting a warning and
-                                       -- closing the connection. Defaults to 3
+      disconnect_timeout_sec?: number    -- How many seconds the client waits for
+                                         -- a disconnect response from the debug
+                                         -- adapter before emitting a warning and
+                                         -- closing the connection. Defaults to 3
 
-      source_filetype?: string         -- The filetype to use for content
-                                       -- retrieved via a source request.
+      source_filetype?: string           -- The filetype to use for content
+                                         -- retrieved via a source request.
+
+      source_map?: table<string, string> -- A table mapping path prefixes from
+                                         -- the editor to the debugger. This is
+                                         -- typically only needed when connecting
+                                         -- to a debugger remotely. Environment
+                                         -- variables are expanded.
 
 
 `dap.adapters.<name>` can also be set to a function which takes three arguments:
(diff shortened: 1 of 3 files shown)


brettmitchelldev:
This is fantastic. I routinely use volume mounts inside containers to develop, and for my use case, it's quite inconvenient to mirror the host's directory structure inside the container.

I've tried this branch out and it solves the problem really well. :rocket:

Perhaps the configuration could be assigned dynamically on a per-config and/or per-attach basis? Might make automation easier. Right now I'm automating assignment by mutating `adapter.options.source_map` directly by reference on attach.

mfussenegger:
> For example, if Nvim is running on a machine where the source tree is located at /a/project/foo and the debugger is running on a separate machine where the source tree is located at /b/project/foo, 

How would this scenario work with gdb and gdbserver ? 
Isn't https://sourceware.org/gdb/current/onlinedocs/gdb.html/Source-Path.html enough for that use-case too?

gpanders:
> > For example, if Nvim is running on a machine where the source tree is located at /a/project/foo and the debugger is running on a separate machine where the source tree is located at /b/project/foo,
> 
> How would this scenario work with gdb and gdbserver ? Isn't https://sourceware.org/gdb/current/onlinedocs/gdb.html/Source-Path.html enough for that use-case too?

If one were running nvim-dap and gdb on the same machine (with the same filesystem layout), with gdb connected to a remote gdbserver instance, then this _may_ work (I no longer have access to the system that I needed this for and tested it with, so cannot say for sure). However that is not always possible. In my case I could not run gdb locally connected to a remote gdbserver instance (it was so slow that it was unusable), but I _could_ run nvim-dap locally connected to gdb running remotely.

I did find this change useful at the time and I used it a bit. But as I said I'm no longer in that position or project so this is no longer directly useful _to me_, so if you want to close it that's fine.

mfussenegger:
>  In my case I could not run gdb locally connected to a remote gdbserver instance (it was so slow that it was unusable), but I could run nvim-dap locally connected to gdb running remotely.

That's interesting to know.

But I think I'll close this for now. The approach would probably have to be changed somewhat to tie it to individual configurations instead of the adapter. That aside, just from a protocol perspective mapping of breakpoint paths can either happen on the client side or on the adapter side - with the adapter potentially having more information to do a meaningful mapping. More importantly there's currently precedence for it to happen on the adapter side.

That said - I'm lukewarm to this and open to arguments. E.g. if someone were to point out a scenario where it can't be done in the adapter, or points to other clients supporting it - that would help change my mind.

PR #1043, opened 2023-09-27
fix: do not append text each time a widgets is opened

### LINKED ISSUE #1042: Calling `toggle` multiple times on a widget causes the new text to get appended instead of replaced

Install any adapter and use any valid nvim-dap configuration.

### Debug adapter version

_No response_

### Steps to Reproduce

1. Create the following mapping to toggle a widget (any widget is fine, not only scopes)

```lua
local scopes = require("dap.ui.widgets").scopes
local sidebar_scopes = require("dap.ui.widgets").sidebar(scopes)
vim.keymap.set("n", "<F5>", function()
  sidebar_scopes.toggle()
end)
```

2. Start a debugging session (with any adapter)
3. Press `<F5>` three times (open it, close it and open it again)

### Expected Result

The widget content should be shown only one time.

### Actual Result

The widget content is shown two times (and if we keep toggling the widgets, each time it gets opened it's content is appended one more time).

TheLeoP:
Fixes #1042.

This is a really naive approach at solving the problem and I'm not sure if this somehow affects other places in the codebase. Since `layer.render` is being called without a parameter for `lnum` and `end_` inside each widget, the render function appends the widget content instead of replacing it (wich does happend when the widget ir refresh. I saw that in that call 0 and -1 were being passed as values for `lnum` and `end_`, so I tried doing the same here and it worked).

### DIFF
--- a/lua/dap/ui.lua
+++ b/lua/dap/ui.lua
@@ -261,6 +261,8 @@ function M.new_tree(opts)
     end,
 
     render = function(layer, value, on_done, lnum, end_)
+      lnum = lnum or 0
+      end_ = end_ or -1
       layer.render({value}, opts.render_parent, nil, lnum, end_)
       if not opts.has_children(value) then
         if on_done then


mfussenegger:
Thanks. I tried it for a bit and seems like there's no unintended side effect

PR #1044, opened 2023-09-28
added option to set widget's position

wybryk-ananasa:
I added option to set widget's position (row and coll).
Setting position also requires `relative`.


mfussenegger:
Wouldn't this also require width and height as part of the `nvim_win_set_config` call?

How are you using this?

mfussenegger:
Closing this for now as there was no feedback.

PR #1060, opened 2023-10-19
Stop always replacing with last scope

NSteinhoff:
`lnum` and `end_` need to be set to `nil` when appending additional scopes to the widget, otherwise the final scope overwrites all previous ones. In the case of `lldb`, where you have 'Local', 'Global', and 'Registers', only the last scope 'Registers' would be drawn to the widget.


NSteinhoff:
This is related to the fix for https://github.com/mfussenegger/nvim-dap/issues/1042

Issue #1052, opened 2023-10-08
REPL: Duplicate newlines when evaluating expressions


wookayin:
### Debug adapter definition and debug configuration

This bug has nothing to do with any particular adapters, but one can use lua adapter for simple reproducibility:

```
  local dap = require('dap')

  dap.configurations.lua = {
    {
      type = 'nlua',
      request = 'attach',
      name = "Attach to running Neovim instance",
    }
  }

  dap.adapters.nlua = function(callback, config)
    callback({
      type = 'server',
      host = config.host or "127.0.0.1",
      port = config.port or 8086
    })
  end

```

### Debug adapter version

https://github.com/jbyuki/one-small-step-for-vimkind

### Steps to Reproduce

1. Open a DAP session and enter the REPL window
2. Type **any expressions** to evaluate, press `<CR>` to enter the line in the REPL

e.g., evaluate a local variable `a`, and constant expressions `1`, `2`, `3`, etc.

### Expected Result

```
dap> a
7
dap> 1
1
dap> 2
2
dap> 3
3
dap> |
```

(`|` means the current cursor)

### Actual Result

Extra "dap> " lines are added before the evaluation result being printed:

```
dap> a
dap>
7
dap> 1
dap>
1
dap> 2
dap>
2
dap> 3
dap>
3
dap> |
```

75a94b8b577a970c0cad25a54a64dc40fc6fa0ea (#889) is the first bad commit after bisecting
/cc @ktakiman @mfussenegger

ktakiman:
Darn it. Thanks for letting me know @wookayin ! I'll give a shot at fixing it - @mfussenegger

ktakiman:
@wookayin  [here is a candidate fix](https://github.com/ktakiman/nvim-dap/tree/fix-repl-empty-prompt) if you'd like to give it a test spin :) 
I'll test it for at least a few days. If everything looks OK, will make a PR.
Thanks!

PR #1063, opened 2023-10-21
Fix repl expressions leave an empty prompt in the middle (#1052)

ktakiman:
Here is a few things I learned along the way and how I approached fixing #1052.
* REPL buffer is based on `buftype=prompt`, where nvim makes sure that there is always a prompt at the end of buffer in insert mode (`dap> ` ).
* When a new command is entered, `exec` function in `repl.lua` is executed (registered via `vim.fn.prompt_setcallback`). nvim does not put the next prompt until callback returns, so, within the callback, new lines can simply be appendend at the end of the buffer. (e.g. `.help` command does not leave an empty prompt in the middle).
* Things get a bit tricky with expression command because when a DAP server request comes back, nvim-dap is aldready outside of the callback, and there is an empty prompt at the end of the buffer. Two options here so not to leave an empty prompt in the middle, 
  1. insert the new line(s) right in front of the empty prompt
  2. overwrite the last line with the new line(s)
* `append` function inside `repl.lua` is updated to apply the first option. `evaluate_handler` now calls this function instead of `layer.render` where possible.
* The second opton is used where `tree.render` is called inside `evaluate_handler`.

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -151,7 +151,7 @@ local function evaluate_handler(err, resp)
   if err then
     local message = utils.fmt_error(err)
     if message then
-      M.append(message)
+      M.append(message, nil, { newline = false })
     end
     return
   end
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -164,18 +164,13 @@ local function evaluate_handler(err, resp)
     -- Appending twice would result in a intermediate "dap> " prompt
     -- To avoid that this eagerly fetches the children; pre-renders the region
     -- and lets tree.render override it
+    local lnum = api.nvim_buf_line_count(repl.buf) - 1
     if spec.has_children(resp) then
       spec.fetch_children(resp, function()
-        local lnum = api.nvim_buf_line_count(repl.buf) - 1
-        local lines = {''}
-        for _ = 1, #resp.variables do
-          table.insert(lines, '')
-        end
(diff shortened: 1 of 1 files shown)


lua/dap/repl.lua
```diff
@@ -164,22 +164,16 @@ local function evaluate_handler(err, resp)
...
     -- and lets tree.render override it
+    local lnum = api.nvim_buf_line_count(repl.buf) - 1
     if spec.has_children(resp) then
       spec.fetch_children(resp, function()
-        local lnum = api.nvim_buf_line_count(repl.buf) - 1
-        local lines = {''}
-        for _ = 1, #resp.variables do
-          table.insert(lines, '')
-        end
-        api.nvim_buf_set_lines(repl.buf, -1, -1, true, lines)
```
ktakiman, line 174: Not 100% sure if it's ok to remove this line, but could not find a case where an empty prompt is left in the buffer without this.

ktakiman:
Tested REPL window with lua (`jbyuki/one-small-step-for-vimkind`), c/c++ (`codelldb`), python(`debugpy`) and javascript (`vscode-node-debug2`) with this update, and so far not seeing any side effect and a blank prompt stopped appearing.

mfussenegger:
Thanks for the great analysis and explanation.

wookayin:
Amazing, thank you @ktakiman!

PR #1064, opened 2023-10-22
Show error messages when the coroutine in trigger_run throws an error

wookayin:
Problem: Any exceptions thrown inside coroutine-wrapped `trigger_run`
inside `dap.run()` are silenced. Dap session won't start silently
without any error messages displayed to users.

Solution: Protect the call of `trigger_run` and show error messages with
stacktrace, if any errors happened.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -480,7 +480,7 @@ function M.run(config, opts)
   if opts.before then
     config = opts.before(config)
   end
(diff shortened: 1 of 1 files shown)


wookayin:
FYI, such errors can happen when adapters are mis-configured; e.g. ENOENT (the adapter command is invalid or can't be spawned)

mfussenegger:
Thanks

Issue #1082, opened 2023-11-04
Support for `command` input type in launch.json


mltoo:
### Problem Statement

VSCode provides support for a `command` type for inputs in launch.json files, which is similar to `pickString`, except a VSCode command is specified to dynamically populate the set of options in the dropdown[1]. I would like to have similar functionality with nvim-dap.

[1] https://code.visualstudio.com/docs/editor/variables-reference#_input-variables

### Possible Solutions

Obviously in neovim we don't have access to VSCode commands, so in order to be able to share launch.json files with VSCode users, I think the best way to do this would be to allow users to specify implementations (e.g. in lua) of VSCode commands for nvim-dap to use.

I have written a PR which allows the user to provide an (optional) additional argument `command_callbacks` to `load_launchjs`. This is a table indexed by names of VSCode commands, each providing a callback function, with the option to provide a 'default' callback in `command_callbacks[1]`, to allow the user to handle parsing/dispatching commands themselves/hand this off to some other plugin. The callback functions are passed the name of the the command called, as well as the `args` section of the input from the launch.json. Callbacks should return a table with strings to appear in the dropdown.

Here is a quick-and-dirty example of how you could use this for a 'pick a unit test to run' launch configuration, in a c++/googletest project. This uses the `shellCommand.execute` VSCode command from [2], which lets you run a shell command by calling a VSCode command:

**Example launch.json**: (The command here simply gets a list of tests from gtest (i.e. TestGroup.TestName), separated by newlines)
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "cppdbg",
            "request": "launch",
            "name": "Run specific test",
            "program": "build/tests",
            "cwd": "${workspaceFolder}",
            "stopOnEntry": false,
            "args": [ "--gtest_filter=${input:pickTest}"]
        }
    ],
    "inputs": [
        {
            "id": "pickTest",
	    "type": "command",
	    "command": "shellCommand.execute",
	    "args": {
		    "command": "build/tests --gtest_list_tests | grep '.*\\.$' | while read group; do build/tests --gtest_list_tests --gtest_filter=\"$group*\" | grep '  .*' | cut -c3- | awk -v g=$group '{print g$1}' ; done"
	    }
        }
    ]
}
```

**Example `command_callbacks` providing a (thrown together and feature-incomplete) implementation of `shellCommand.execute`:**

```lua
local command_callbacks = {}
command_callbacks['shellCommand.execute'] =  function(command, args)
[...]

PR #1083, opened 2023-11-04
Add support for `command` input type in launch.json

mltoo:
See #1082.


mfussenegger:
I'm not yet sure how to deal with this.

It would be great if at least some of the basics just work out of the box, but looking at what kind of commands exist it seems that that's basically _everything_ one can do in vscode. That's not feasible to include and including a subset would make it harder to draw a clear line - although not impossible.

So making this extensible seems like a good call, but I think I'd rather go in a even more general direction to also account for https://github.com/mfussenegger/nvim-dap/issues/720
That means some sort of extensible configuration pre-processing mechanism

Issue #1062, opened 2023-10-21
More variable formatting options and/or extension points 


mfussenegger:
### Problem Statement

This is more of a collection of a few pain points in regards to the presentation and functionality related to variables.

Mostly a brain dump for now.

### long text

The display of variables can be quite hard to read if it is a structure with a
very long string representation as you get something like:

```
looooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooonnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnnngggggggggggggggggggggggggggggggggggggggggggggggg tttttttttttttttttttttttttttttttttttttttteeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeexxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxtttttttttttttttttttttttttttttttttttttttttttttttttttttt
property1: foo
property2: bar
property3: ..
[...]
```

### Lazy variables

Some debug adapters make heavy use of the `lazy` presentation hint, which requires an extra expansion step.

For example, with java-debug:

```
  settings: ArrayList@50 size=1
```

Expanding that shows:

```
    0: Assignment@66
```

And on a second expansion:

```
    0: Assignment@66 "Assignment{column='stats.jobs_log_size', expressions=[1024]}"
```

This extra step can be a bit annoying, especially for cases where you want to have an overview over a collection. (Like when you view a stacktrace)

### Formatters

Sometimes you stop at a breakpoint and view the contents of a variable and then want to create a test-case with that data. It would be cool if there were some way to control how the variable is formatted.

E.g. a "copy as Lua/JSON/Python/.."
[...]

PR #1089, opened 2023-11-12
Make dap.repl.evaluate_handler public

wookayin:
This allows users to have a minimal configuration point for customizing
dap-repl's behavior on printing variables or expressions.

(One possible approach/workaround for #1062)

### DIFF
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -147,7 +147,7 @@ local function print_commands()
 end
 
 
-local function evaluate_handler(err, resp)
+function M.evaluate_handler(err, resp)
   if err then
     local message = utils.fmt_error(err)
     if message then
@@ -290,7 +290,7 @@ function execute(text)
     local args = string.sub(text, string.len(command)+2)
     M.commands.custom_commands[command](args)
   else
-    session:evaluate(text, evaluate_handler)
+    session:evaluate(text, M.evaluate_handler)
   end
 end
 


mfussenegger:
Thanks but I'd rather keep that prevent. Exposing it might get in the way latter when adding proper formatting extensions.

If you you really want to override the full handler, you could instead define a custom command:

```lua
  repl.commands.custom_commands = {
    [".p"] = function(text)
      local session = dap.session()
      if not session then
        return
      end
      session:evaluate(text, function(err, resp)
        vim.print(err or "no error")
        vim.print(resp or "no response")
      end)
    end,
  }
```

wookayin:
I understand that you don't want this merged. However, what I wanted to override is the `evaluate_handler` without any custom commands (for #737 #1062, etc.), so custom commands would not work for me.

mfussenegger:
> However, what I wanted to override is the evaluate_handler without any custom commands (for https://github.com/mfussenegger/nvim-dap/issues/737 https://github.com/mfussenegger/nvim-dap/issues/1062, etc.), evaluation of any arbitrary expressions in the REPL; so custom commands would not work for me.

The snippet I posted allows to evaluate arbitrary expressions. `.p foobar` would evaluate `foobar`.

wookayin:
I'd like to have `foobar` working instead of typing `.p foobar`.

```
dap> foobar
"foobar"
```

PR #1090, opened 2023-11-12
Allow filetype in options to dap.continue and dap.run

wookayin:
`dap.run { filetype = ... }` will start a DAP session with the specified
filetype, with the configuration `dap.configurations[filetype]`.

NOTE: Currently, no matter what the filetype of the current buffer is picked.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -411,7 +411,7 @@ end
 
 
 local function select_config_and_run(opts)
-  local filetype = vim.bo.filetype
+  local filetype = opts.filetype or vim.bo.filetype
   local configurations = M.configurations[filetype] or {}
   assert(
     vim.tbl_islist(configurations),
@@ -427,7 +427,7 @@ local function select_config_and_run(opts)
     return
   end
   opts = opts or {}
-  opts.filetype = opts.filetype or filetype
+  opts.filetype = filetype
   lazy.ui.pick_if_many(
     configurations,
     "Configuration: ",
@@ -903,7 +903,7 @@ function M.run_to_cursor()
 end
 
 
----@param opts? {new?: boolean}
+---@param opts? {new?: boolean, filetype?: string}
 function M.continue(opts)
   if not session then
     session = first_stopped_session()


wookayin:
@mfussenegger any opinions?

mfussenegger:
I have in mind to add some sort of extensible configuration provider mechanism but I don't have a clear picture yet how that will look like. But I suspect it will move a bit away from the strict filetype coupling so
I'm not sure if adding a filetype parameter will then still fit. There might be a more general filtering mechanism/table parameter, given that each provider implementation may take different options.

I can imagine that the two existing ways to define configurations (dap.configurations, launch.json) will then just become bundled configuration provider implementations.
The vscode launch.json entries wouldn't get added/loaded into `dap .configurations` and you'd always get all entries listed in the file.

wookayin:
The current strict filetype coupling is a limitation that bothers me; it makes remote attaching especially difficult (e.g. run "cpp" or "lua" configuration when on an "empty" buffer, etc.) But I think this might be a good and simple solution until a complex configuration or provider mechanism is landed.

ckyrouac:
I was about to make a PR with something similar. This works for me to select any run configuration from any filetype:

```
local function flatten_dap_configurations()
  local configurations = require('dap').configurations
  local flat_configs = {}

  for type, configs in pairs(configurations) do
    for _, conf in ipairs(configs) do
      conf.file_type = type
      table.insert(flat_configs, conf)
    end
  end

  return flat_configs
end

local function select_config_and_run(opts)
  local configurations = flatten_dap_configurations()
  opts = opts or {}
  require('dap.ui').pick_one(
    configurations,
    "Configuration: ",
    function(i) return i.file_type .. ": " .. i.name end,
    function(configuration)
      if configuration then
        require('dap').run(configuration, opts)
      else
        vim.notify('No configuration selected', vim.log.levels.INFO)
      end
    end
  )
end
vim.keymap.set("n", "<F10>", select_config_and_run, { silent = true, desc = 'DAP Select Config' })
```

mfussenegger:
Can you check if https://github.com/mfussenegger/nvim-dap/pull/1237 would solve this requirement?

mfussenegger:
Closing this for now but please let me know if your use cases can't be covered with the provider model

Issue #1096, opened 2023-11-19
Filtering on `dap.util.get_processes`


NicholasMata:
### Problem Statement

I recently decided to make a language specific extension. Don't worry I know the README mentions that isn't the goal of this library. 

I did notice there are some nice utility functions in `dap.utils`

`pick_process` was modified in  #922 to allow filtering, which is very useful. In building my extension I noticed it would also be useful to add filtering on `get_processes` (I know I can just filter it myself which is what I am currently doing). However I think it would make sense for it exist in this library.

### Possible Solutions

I think a good solution would be to move the filtering that currently exists in `pick_process` into `get_processes`. I think this makes the most sense because `pick_process` internally calls `get_processes` and then filters the results within itself. 

### Considered Alternatives

To leave filtering out of this library, which is the way it currently exists

PR #1097, opened 2023-11-19
Moved filtering logic from `pick_process` to `get_processes`

NicholasMata:
This implements a feature request I recently made #1096. If you believe it makes sense then here is a PR which implements the proposed change.


mfussenegger:
Thanks

PR #1121, opened 2023-12-26
Show last message in statusline

NSteinhoff:
If there are pending messages, they would previously be displayed one by one in the status line on each redraw (typically when moving the cursor). Instead, the last message is shown always.

This may have been the intended behavior, but the documentation for `dap.status()` does not mention it. If so, please close this PR.

### DIFF
--- a/lua/dap/progress.lua
+++ b/lua/dap/progress.lua
@@ -48,13 +48,12 @@ function M.status()
   if not session then
     return ''
   end
-  local msg = M.poll_msg() or last_msg
-  if msg then
-    last_msg = msg
-    return msg
-  else
-    return ''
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks for the PR but that's working as intended. If you want to update the status line immediately after a new message you can add a autocmd on `User DapProgressUpdate` to refresh the status:

```lua
vim.api.nvim_create_autocmd("User", {
  group = vim.api.nvim_create_augroup("dap-status", { clear = true }),
  pattern = "DapProgressUpdate",
  command = "redrawstatus"
})
```

NSteinhoff:
Awesome, thanks!

PR #1124, opened 2023-12-28
Prioritize local debugging method

TroySigX:
`load_launchjs` is  usually called after setting up `dap.configurations`, this leads to the local settings being put at the bottom of the methods selection list. But most of the time, user would prefer it to be at the top for convenience.

### DIFF
--- a/lua/dap/ext/vscode.lua
+++ b/lua/dap/ext/vscode.lua
@@ -204,7 +204,7 @@ function M.load_launchjs(path, type_to_filetypes)
           table.remove(dap_configurations, i)
         end
       end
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks for the PR.

One downside of this PR is that users could no longer get the old behavior back if they wanted to.

On the other hand, without this PR it's possible to call `load_launchjs` before adding manual dap.configuration entries to get this new behavior.

Because of that I'd keep this open a while longer to get some more comments / votes

mfussenegger:
Didn't gain any votes so I'm closing this.

PR #1129, opened 2024-01-06
feat: add sign for hit condition breakpoints

jonathan-elize:
I noticed there wasn't a sign for breakpoints with a hit condition. This PR is to add one for them.


mfussenegger:
I'm a bit warry of adding more signs given that some of the options can also be used in combination (you can have conditional log points for example)

I might revisit this later but for now I'm not merging this.

PR #1190, opened 2024-04-14
feat(utils): Implement pick_files()

### LINKED ISSUE #1130: Add picker for executables/files
### Problem Statement

You provide a nice picker to select a process to get the pid. It would be great to have the same available to select programs/executables to debug.

I'm currently using:

```lua
        program = function()
            local path = vim.fn.input({
                prompt = 'Path to executable: ',
                default = vim.fn.getcwd() .. '/',
                completion = 'file',
            })

            return (path and path ~= '') and path or dap.ABORT
        end,
        args = function()
            local args_str = vim.fn.input({
                prompt = 'Arguments: ',
            })
            return vim.split(args_str, ' +')
        end,
```

### Possible Solutions

It would be nice if you would have a picker searching for executables in the directory.

```lua
    program = require('dap.utils').pick_file({ executables_only = true })
```
It could also be used maybe with a filter:

```lua
    program = require('dap.utils').pick_file({ filter = '*.py' })
```

### Considered Alternatives

_No response_

cryptomilk:
This implements a file picker, to pick the file you want to execute with the debugger.

Fixes #1130

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -1265,7 +1265,7 @@ pick_process({opts})                                    *dap.utils.pick_process*
 pick_file({opts})                                         *dap.utils.pick_file*
     Show a prompt to select a file.
     Returns the path to the selected file.
-    Requires nvim 0.11
+    Requires nvim 0.10+
 
     Parameters:
       {opts} optional table with:
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -1260,6 +1260,34 @@ pick_process({opts})                                    *dap.utils.pick_process*
       filter = function(proc) return vim.endswith(proc.name, "sway") end
     })
 <
+
+
+pick_file({opts})                                         *dap.utils.pick_file*
+    Show a prompt to select a file.
+    Returns the path to the selected file.
+    Requires nvim 0.11
+
+    Parameters:
+      {opts} optional table with:
+
+              - filter? string|fun: A |lua-pattern| or a function to filter the
+                files.
+                If a function the parameter is a string and it must return a
+                boolean. Matches are included.
+
+              - executables? boolean: Show only executables. Defaults
+                to true
+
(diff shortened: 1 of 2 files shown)


cryptomilk:
It looks like this with telescope dap:

image

cryptomilk:
Ping?

mfussenegger:
Pushed a couple of changes:

- Uses vim.fs.dir instead of find/bfs to make it platform independent (but requires nvim 0.11)
- Default executables to true as I think that's the common use case and it allows to use it like `program = require("dap.utils").pick_file,` in a configuration entry, without wrapping it in another function to provide the arguments.
- Simplified the coroutine logic, given that `pick_one` is already coroutine aware and the callback is optional.
- Fixed the filter example. It is a lua-pattern, not a glob pattern.

Open questions to me:

- Could re-add `find` as fallback for nvim 0.10 and earlier.
- Could change the filter from a lua pattern to a glob via `vim.glob.to_lpeg()` - might be more intuitive given that it is about file paths, but could also be confusing if it is different to `pick_process`

cryptomilk:
A fallback for nvim 0.10 would be nice as it just has been released :-)

mfussenegger:
> A fallback for nvim 0.10 would be nice as it just has been released :-)

Ah, I got the numbers mixed. It requires 0.10. The fallback would make it work in 0.9 or earlier.

cryptomilk:
> Ah, I got the numbers mixed. It requires 0.10. The fallback would make it work in 0.9 or earlier.

Ah, then I would argue new feature new nvim version required :-)

mfussenegger:
Decided to add in the `find` fallback (but left `bfs` out). I also changed the result to return dap.ABORT, so it doesn't attempt to start the debug session if one cancels the dialog.

Issue #1220, opened 2024-05-23
pick_files() doesn't work on neovim 0.10


cryptomilk:
### Debug adapter definition and debug configuration

pick_files only gives me an empty list:

```lua
local ok, dap = pcall(require, 'dap')
if not ok then
    return
end

local dap_utils = require('dap.utils')

local cwd = vim.fn.getcwd()
local is_samba = cwd:find('samba')

local path = cwd
if is_samba then
    path = cwd .. '/bin'
end

local exec_opts = {
    path = path,
    executables = true,

    filter = function(exec)
        -- Filter out shared libraries
        return not exec:match('%.so([.0-9]*)')
    end,
}

--
-- See
-- https://sourceware.org/gdb/current/onlinedocs/gdb.html/Interpreters.html
-- https://sourceware.org/gdb/current/onlinedocs/gdb.html/Debugger-Adapter-Protocol.html
dap.adapters.gdb = {
    id = 'gdb',
    type = 'executable',
    command = 'gdb',
    args = { '--quiet', '--interpreter=dap' },
    --[[
    args = {
        '-iex',
        'set debug dap-log-file /tmp/gdb-dap.log',
        '--quiet',
        '--interpreter=dap',
    },
]

dap.configurations.c = {
    {
        name = 'Run executable (GDB)',
        type = 'gdb',
        request = 'launch',
        -- This requires special handling of 'run_last', see
        -- https://github.com/mfussenegger/nvim-dap/issues/1025#issuecomment-1695852355
        program = function()
            return dap_utils.pick_file(exec_opts)
        end,
    },
    {
        name = 'Run executable with arguments (GDB)',
        type = 'gdb',
        request = 'launch',
        -- This requires special handling of 'run_last', see
        -- https://github.com/mfussenegger/nvim-dap/issues/1025#issuecomment-1695852355
        program = function()
            return dap_utils.pick_file(exec_opts)
        end,
        args = function()
            local args_str = vim.fn.input({
                prompt = 'Execeutable arguments: ',
            })
            return vim.split(args_str, ' +')
        end,
    },
    {
        name = 'Attach to process (GDB)',
        type = 'gdb',
        request = 'attach',
        processId = require('dap.utils').pick_process,
    },
}
```

### Debug adapter version

master

### Steps to Reproduce

1. Open C file
2. Set Breakpoint
3. Try to run it with gdb config
4. Get executable file list

### Expected Result

Get a executable file list

### Actual Result

Empty list

[...]

PR #1221, opened 2024-05-23
Handle paths relative to custom path in utils.pick_file

mfussenegger:
Should fix https://github.com/mfussenegger/nvim-dap/issues/1220

### DIFF
--- a/lua/dap/utils.lua
+++ b/lua/dap/utils.lua
@@ -220,8 +220,11 @@ local function get_files(path, opts)
   if vim.fs.dir then
     local files = {}
     for name, type in vim.fs.dir(path, { depth = 50 }) do
-      if type == "file" and filter(name) then
-        table.insert(files, name)
+      if type == "file" then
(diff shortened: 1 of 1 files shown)


cryptomilk:
Unfortunatly this doesn't fix it :-(

cryptomilk:
Oh. the directory is wrong:

```
stat on default/source3/lib/netapi/examples/localgroup/localgroup_getmembers                                                                                                        
no stat
```

I pass /home/asn/workspace/prj/oss/samba/bin too it. So it thinks this is the current working directory, but the cwd is `/home/asn/workspace/prj/oss/samba`

It should be `bin/default/source3/lib/netapi/examples/localgroup/localgroup_getmembers `

mfussenegger:
Right, I should've used the joined path for the filter too :)

Can you check the last fixup?

cryptomilk:
This works! Yay :-)

Issue #1226, opened 2024-05-26
rfc: configuration providers


mfussenegger:
### Problem Statement

Currently you always need to be in a file with the right filetype to start a debug session.

This works well enough for many work patterns, but on some type of projects it's common that you're in another buffer. E.g. you might want to start `django` while editing a `HTML` file. Or you're in the REPL and want to start a debug session again.

### Possible Solutions

Showing _all_ global configuration options is probably too much, but a compromise could be to:

1. Show all local configurations (from `.vscode/launch.json`)
2. Show global configurations related to the filetype as is.
3. Provide a `provider.configs` extension points for additional sources. (E.g. some language servers/plugins provide functionality to discover entry points)

Always doing 1) could also remove the need to specify a filetype/debug-adapter-type mapping as part of the `load_launchjs` call.

1) and 2) would be implemented using the provider/configurations extension points, that means people could also unregister or replace them.

For 3) I'm thinking of something like:

```
local dap = require("dap")
dap.provider.configs[plugin_name] = function()
    -- can return a list of configurations, or a thread that will yield the list
end
```

(I'm also thinking of other types of providers, like variable formatting or `yank-as-<format>` functionality - hence the `provider` namespace.  Or `dap.ext.`?)

I'd like to get some feedback on this.

mfussenegger:
Implemented with https://github.com/mfussenegger/nvim-dap/pull/1237

Issue #1104, opened 2023-11-30
VS Code launch JSON type


Otard95:
### Debug adapter definition and debug configuration

Not sure if it's entirely correct to call this a bug, but I feel like it's more than just a discussion.

Adapter installed via `mason-nvim-dap`.

```lua
  require 'mason-nvim-dap'.setup {
    ensure_installed = {
      'js',
      'php'
    },
  }
```

Configured like:
```lua
  require 'dap.ext.vscode'.load_launchjs('.vscode/launch.json', {
    node = { 'typescript', 'javascript' },
  })
  dap.adapters["node"] = {
    type = 'server',
    port = 8123,
    executable = {
      command = 'js-debug-adapter',
      args = { '8123' },
    },
    options = {
      max_retries = 40,
    },
  }
```

### Steps to Reproduce

Any basic configuration should yield the same result.

### Expected Result

I'd like to see `"type": "node",` in launch.json work. Considering that VS Code expects this, though not required.

![image](https://github.com/mfussenegger/nvim-dap/assets/7838337/c27e31b6-1ac4-4a03-b616-696108fb1b4f)

### Actual Result

The debugger fails with `Error: Unknown config`.

Considering VS Code want's us to use `node` instead of `pwa-node` maybe `require 'dap.ext.vscode'.load_launchjs(...)` should change the `node` type to `pwa-node`. Of course this creates a little bit of a hidden disconnect between the type in the configuration and adapter, but considering VS Code itself urges us not to use `pwa-node`, it might be the better option?

mfussenegger:
As far as I can tell the pwa prefixes are still required in the launch configuration: 

https://github.com/microsoft/vscode-js-debug/blob/e51f7cc6ccb3f44b328e386f28109f305bc5741d/src/common/contributionUtils.ts#L65-L71 

I suspect their client extension might map node to pwa-node somewhere to make it work with just `node`. 
I'm not adding debug adapter specific logic to nvim-dap. This would have to change in the adapter

Otard95:
I agree, you shouldn't. But I feel we may have talked past each other. I'm not suggesting a change to how nvim-dap talks to the adapter.

But from what I understand `load_launchjs` is a util to allow you to use a VS Code launch configuration with nvim-dap, and if so would it not make sense to have that util, and just that util, handle a configuration in the format VS Code wants it? That is it should translate a `launch.json` to what the debug adapter expects?

[...]

PR #1232, opened 2024-05-29
Add ability to specify what adapter a type corresponds to from launch.json

jonathan-elize:
I use [nvim-dap-vscode-js](https://github.com/mxsdev/nvim-dap-vscode-js) to debug javascript and typescript files. It sets up adapters automatically with various names such as `pwa-node`.

In launch.json files, the "type" that is put for debugging javascript and typescript files I've found to usually be set to "node". Since right now the type is used to find the adapter by name in nvim-dap, trying to debug using these configs is failing by default since I don't have an adapter named "node". I'm creating this PR to be able to map type from the launch.json configuration to the adapter you would want to use in Neovim.

In my case, with this change I can now do this to load debug options from launch.json and run them successfully.

```lua
  local vs_map = { node = { "javascript", "typescript" } }
  local adapter_map = { node = "pwa-node" }
  require("dap.ext.vscode").load_launchjs(nil, vs_map, adapter_map)
```

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -333,7 +333,7 @@ function to support json5 features like trailing comma.
 (One json5 parser implementation is https://github.com/Joakker/lua-json5)
 
 
-load_launchjs({path}, {type_to_filetypes})        *dap.ext.vscode.load_launchjs*
+load_launchjs({path}, {type_to_filetypes}, {type_to_adapter})        *dap.ext.vscode.load_launchjs*
 
   Parses a JSON file at {path} and adds the entries to `dap.configurations`.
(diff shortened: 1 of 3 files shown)


mfussenegger:
Thanks for the PR but there is already a solution for this. See https://github.com/mfussenegger/nvim-dap/issues/1104#issuecomment-1835734572

I'm also thinking of always loading launch.json implicitly when starting a debug session and drop the filetype-type mapping part. Having more config options for it would counter that plan. See https://github.com/mfussenegger/nvim-dap/issues/1226 for details

PR #1244, opened 2024-05-31
feat: Add neoconf.nvim config provider

AThePeanut4:
This adds a config provider for https://github.com/folke/neoconf.nvim, providing a more "Neovim-native" alternative to `.vscode/launch.json` files.

I added some hacks locally quite a while ago to add rudimentary support, and had actually got around to cleaning it up last week and was going to make a PR. The reason why I didn't was because I was thinking of proposing something basically identical to #1237, to avoid having to make a bunch of changes directly to `dap.lua`. So @mfussenegger thanks for implementing that and reading my mind :)

What I've done is add a provider that does the exact same thing as the vscode one, the only difference being that it takes its input from neoconf. The common configuration and input logic I moved out of `vscode.lua` into `utils.lua` - not sure if the file name fits?

The plugin also registers itself with neoconf to provide a config schema, which allows for nice autocomplete in the `neoconf.json` file.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -301,6 +301,13 @@ providers.configs["dap.launch.json"] = function()
   return ok and configs or {}
 end
 
+providers.configs["dap.neoconf"] = function(bufnr)
+  local ok, configs = pcall(function()
+    return require("dap.ext.neoconf").getconfigs(bufnr)
+  end)
+  return ok and configs or {}
+end
+
 do
   local function eval_option(option)
     if type(option) == 'function' then
@@ -1209,5 +1216,6 @@ api.nvim_create_autocmd("ExitPre", {
   end
 })
 
+require("dap.ext.neoconf").register()
 
 return M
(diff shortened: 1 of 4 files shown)


mfussenegger:
Thanks for the PR but I'm sorry I'm not going to merge this.

 The point of the provider model is that other plugins can extend it, not to add more built-in providers.

mfussenegger:
To add to that. I also don't fully understand the intention behind this. 
The PR re-uses pretty much all of the launch.json code, which means the format is 1:1 the same? 

What's making this more neovim-native and why is this desirable?

The idea behind re-using vscode/launch.json instead of inventing a custom format was that you can share the configuration files between team members using different editors. The configurations are typically 100% project specific, and not in any way personalized.
This seems to go against that and fosters duplication?

AThePeanut4:
Fair, I see your point. I hadn't really thought about sharing configurations.

The advantage to neoconf over `.vscode/launch.json` I see as mostly just being able to keep all project-local "IDE" configuration in one single file. A Neovim `.idea` folder in a sense, but as a single file.

PR #1273, opened 2024-06-26
Remove executable from adapter when request type is attach

jsternberg:
Related to https://github.com/leoluz/nvim-dap-go/issues/35. When attempting to attach to a remote dlv process, the nvim-dap-go plugin would attempt to launch an executable on the same port it was supposed to connect to. This would fail and would cause the plugin not to attach to the existingøØrt.

This is because nvim-dap-go switched to using the included adapter and the adapter does not check if the request is launch or attach before attempting to launch the dlv process.

This is fixed by removing executable from the adapter if the config uses attach since an attach request shouldn't be launching an executable anyway.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -1122,6 +1122,15 @@ function M.attach(adapter, config, opts)
     return
   end
   assert(adapter.port, 'Adapter used with attach must have a port property')
+
+  -- Clear executable from adapter when the request is to attach.
+  -- Spawning a new executable is likely the wrong thing to do.
+  if config.request == 'attach' and adapter.executable then
+    adapter = vim.deepcopy(adapter)
+    ---@diagnostic disable-next-line: inject-field
+    adapter.executable = nil
+  end
+
   local s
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks for the PR but this is the wrong fix and would end up breaking other adapters.

Adapters tend to treat remote access in one of two ways:

- The remote end speaks DAP natively. In this case the client can connect to the remote end directly via a `server` adapter definition. It looks like this is the case for delve?
- The remote end doesn't speak DAP natively and a local adapter executable is still needed. Afaik an example of this is codelldb. This PR would break this case.

The proper fix would be to change the adapter configuration that nvim-dap-go is making, similar to the one used in nvim-dap-python:

https://github.com/mfussenegger/nvim-dap-python/blob/d777c2b32ed39f61209c09bede28d7491621a631/lua/dap-python.lua#L208-L234

jsternberg:
Thanks for the direction. I'll do that instead.

PR #1281, opened 2024-07-17
Add support for sessionless expression interpreters

mfussenegger:
This is just a prototype. I might end up not merging this.

[demo](https://zignar.net/assets/files/c6144a41526e81b82d9cac39901782a4140d3f14adec3c8b8061cb1028e700ff.webm)


mfussenegger:
Decided against including this, but created https://github.com/mfussenegger/nluarepl which gives (almost) the same functionality.

PR #1285, opened 2024-07-19
Fix only general error message is shown when launch.json is invalid

ADIX7:
When launch.json has invalid encoding/bom launch.json can not be parsed, but it seems that assert does not catches it, so there is a general error message about can not parsing the json, instead of the message that there is some kind of problem with launch.json

To reproduce (tested on Linux):
```
:set ff=unix
:set bomb
```

This PR fixes this behaviour. Currently it also shows the whole launch.json, so the `<feff>` is very telling. If it is too much I can revert to the original message. Maybe extending it with a hint that BOM might be the source of the error.

### DIFF
--- a/lua/dap/ext/vscode.lua
+++ b/lua/dap/ext/vscode.lua
@@ -147,8 +147,7 @@ end
 function M._load_json(jsonstr)
   local ok, data = pcall(M.json_decode, jsonstr)
   if not ok then
-    notify("Error parsing launch.json: " .. jsonstr, vim.log.levels.ERROR)
-    return {}
+    error("Error parsing launch.json: " .. data)
   end
 
   local inputs = create_inputs(data.inputs or {})
--- a/lua/dap/ext/vscode.lua
+++ b/lua/dap/ext/vscode.lua
@@ -145,7 +145,12 @@ end
 
 
 function M._load_json(jsonstr)
-  local data = assert(M.json_decode(jsonstr), "launch.json must contain a JSON object")
+  local ok, data = pcall(M.json_decode, jsonstr)
+  if not ok then
+    notify("Error parsing launch.json: " .. jsonstr, vim.log.levels.ERROR)
+    return {}
+  end
+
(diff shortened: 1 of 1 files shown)


lua/dap/ext/vscode.lua
```diff
@@ -145,7 +145,12 @@ end
 
 
 function M._load_json(jsonstr)
-  local data = assert(M.json_decode(jsonstr), "launch.json must contain a JSON object")
+  local ok, data = pcall(M.json_decode, jsonstr)
+  if not ok then
+    notify("Error parsing launch.json: " .. jsonstr, vim.log.levels.ERROR)
+    return {}
+  end
+
```
mfussenegger, line 153:
```suggestion
  local ok, data = pcall(M.json_decode, jsonstr)
  if not ok then
    error("Error parsing launch.json: " .. data)
  end

```

- Including the whole content is imho too much. `data` will contain the error message if pcall failed which should give a good enough hint ("Expected value but found invalid token at character 1").
- This must still raise an error to not break error suppression in other places. E.g.:

https://github.com/mfussenegger/nvim-dap/blob/bc03b83c94d0375145ff5ac6a6dcf28c1241e06f/lua/dap.lua#L299-L302

mfussenegger (commented):
Thanks for the PR

lua/dap/ext/vscode.lua
```diff
@@ -145,7 +145,11 @@ end
 
 
 function M._load_json(jsonstr)
-  local data = assert(M.json_decode(jsonstr), "launch.json must contain a JSON object")
+  local ok, data = pcall(M.json_decode, jsonstr)
+  if not ok then
+    error("Error parsing launch.json: " .. data)
+  end
+
```
mfussenegger, line 152:
```suggestion
  local ok, data = pcall(M.json_decode, jsonstr)
  if not ok then
    error("Error parsing launch.json: " .. data)
  end
  assert(type(data) == "table", "launch.json must contain a JSON object")
```

PR #1295, opened 2024-07-27
Add utils.pick_process_by_name()

barreiroleo:
Hi!

My approach to manage the PID picker.
1) Enter the process name to attach
2) pgrep will retrieve the list off matching process and the user can select one of them
3) If none results was found, the user can input one by hand

Unfortunately it doesn't support windows because I don't have a proper setup to test it.

Config prompt:
imagen
Process name prompt:
imagen
Select PID prompt:
imagen
Manual PID prompt:
imagen
Attached:
imagen

### DIFF
--- a/lua/dap/utils.lua
+++ b/lua/dap/utils.lua
@@ -302,4 +302,45 @@ function M.pick_file(opts)
 end
 
 
+--- Show a prompt to input a process name and select one of the the matching PIDs.
+--- Returns Process ID to attach the adapter.
+--- Requires a pgrep executable and a unix system.
+---
+--- <pre>
+-- {
+--     name = '[codelldb] Attach to PID',
+--     type = 'codelldb',
+--     request = 'attach',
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks for the PR but the behavior of this is imho too similar to `pick_process({ filter = vim.fn.input(...) })` to warrant an extra function so I'd rather not merge it given that it's also pretty simple to define outside if really needed.

barreiroleo:
Yeeah, I didn't realize about the filter, my bad 😅

For future references, or if you want to push it into the docs, an equivalent example:
```
{
        name = '[codelldb] Attach to PID',
        type = 'codelldb',
        request = 'attach',
        cwd = '${workspaceFolder}',
        pid = function()
            return require("dap.utils").pick_process({ filter = vim.fn.input("Process name: ") })
        end
}
```

Thanks!

Issue #956, opened 2023-05-21
Debug adpater reported a frame at line 74, cursor position outside buffer.


akashdiphazra:
### Debug adapter definition and debug configuration

C/C++ Debugger Doesn't exit properly.
```lua
local dap = require('dap')
dap.adapters.cppdbg = {
  id = 'cppdbg',
  type = 'executable',
  command = '/usr/local/nvim-dap-adapters/cpp-tools/extension/debugAdapters/bin/OpenDebugAD7',
}
dap.configurations.cpp = {
  {
    name = "vscode-cpptools",
    type = "cppdbg",
    request = "launch",
    program = function()
      return vim.fn.input('Path to executable: ', vim.fn.getcwd() .. '/', 'file')
    end,
    cwd = '${workspaceFolder}',
    launchCompleteCommand = "exec-run",
    stopAtEntry = false,
    miDebuggerArgs = "-x ./.gdbinit",
    setupCommands = {
      {
        text = '-enable-pretty-printing',
        description = 'enable pretty printing',
        ignoreFailures = false
      },
    },
  },
  {
    name = 'gdbserver',
    type = 'cppdbg',
    request = 'launch',
    MIMode = 'gdb',
    cwd = '${workspaceFolder}',
    miDebuggerPath = '/usr/bin/gdb',
    miDebuggerArgs = "-x ./.gdbinit",
    program = function()
      return vim.fn.input('Path to executable: ', vim.fn.getcwd() .. '/', 'file')
    end,
  },
}


```

### Debug adapter version

0.6.0

### Expected Result

Debugger should stop and exit.

### Actual Result

[Screencast from 2023-05-21 22-23-51_trimmed.webm](https://github.com/mfussenegger/nvim-dap/assets/61003681/d6265ab2-0a31-42fe-a2aa-de6cbf888a64)

mfussenegger:
Part of the idea behind reproduction steps is that one can follow them to reproduce the problem.

akashdiphazra:
### I tried to skip stl and other .h files , so that it's doesn't step into unnecessary implementation, It's disturbing and overwhelming during debugging process because unnecessary files open randomly. Unlike vscode and visual studio debugger JusMyCode feature never works or not fully implement yet even on windows also , it tried many times but failed. So the last option was using .gdbinit file.
my .gdbinit file

````shell
skip -gfi /usr/include/c++/12/*/*/*
skip -gfi /usr/include/c++/12/*/*
[...]

PR #1315, opened 2024-08-13
Handle disassembly buffer

bogdan-nikitin:
Fix https://github.com/mfussenegger/nvim-dap/issues/956#issuecomment-1584883229

The disassembly buffer is of the "nofile" type, which has not been handled. This leads to the error "cursor position outside the buffer" when more than two panes opened.

This error can be reproduced using [this Dockerfile](https://github.com/bogdan-nikitin/nvim-dap-956/blob/main/Dockerfile):
1. Run container interactively. Neovim opens with an example file with a breakpoint set
2. Type `:DapContinue`
3. Type `:DapStepInto` few times
4. Assembly will appear in both panes

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -410,7 +410,9 @@ local function jump_to_location(bufnr, line, column, switchbuf, filetype)
   local switchbuf_fn = {}
 
   function switchbuf_fn.uselast()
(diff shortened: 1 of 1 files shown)


bogdan-nikitin:
Now I'm handling the disassembly buffer by checking the Vim variable `b:dap_source_buf`

mfussenegger:
> Now I'm handling the disassembly buffer by checking the Vim variable `b:dap_source_buf`

Thanks

PR #1318, opened 2024-08-16
Fix run in external terminal not propagating environment variables to program

### LINKED ISSUE #1317: Run in external terminal does not propagate environment variables correctly

I am writing a debug adapter and noticed that while the integrated terminal correctly propagates environment variables, the external terminal does not.

### Debug adapter version

2b428ff

### Steps to Reproduce

1. Use any debugger supporting passing environment variables and running in external terminal
2. Write a program which prints a sample variable
3. Set the sample variable to a text

### Expected Result

The sample variable gets printed

### Actual Result

No sample variable is set

sewbacca:
This PR fixes #1317
and correctly passes environment variables to the program

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -112,15 +112,23 @@ local function coresume(co)
 end
 
 
-local function launch_external_terminal(terminal, args)
+local function launch_external_terminal(env, terminal, args)
   local handle
   local pid_or_err
   local full_args = {}
   vim.list_extend(full_args, terminal.args or {})
   vim.list_extend(full_args, args)
+  local env_formatted = {}
+  -- Copy environment, prefer vars set by client
+  for k, v in pairs(env and vim.tbl_extend('keep', env, vim.fn.environ()) or {}) do
+    if k:find "^[^=]*$" then -- correct variable?
+      env_formatted[#env_formatted+1] = k.."="..tostring(v)
+    end
+  end
   local opts = {
     args = full_args,
-    detached = true
+    detached = true,
+    env = env_formatted,
   }
   handle, pid_or_err = uv.spawn(terminal.command, opts, function(code)
     if handle then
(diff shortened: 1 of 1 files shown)


lua/dap/session.lua
```diff
@@ -112,15 +112,22 @@ local function coresume(co)
...
   local pid_or_err
   local full_args = {}
   vim.list_extend(full_args, terminal.args or {})
   vim.list_extend(full_args, args)
+  local env_formatted = {}
+  for k, v in pairs(env or {}) do
+    if k:find "^[^=]*$" then -- correct variable?
+      env_formatted[#env_formatted+1] = k.."="..tostring(v)
+    end
+  end
```
mfussenegger, line 126: I think if `env` is set then `spawn` uses only these variables, without inheriting/merging any from the parent process.
I suspect this would need to include at least `PATH` but might be safer to copy the entire `vim.fn.environ()` and extend it based on `env`

sewbacca: Fixed using `vim.extend_tbl` in b8e54ac91cae94df6c609d986a7776e4ffc744ca

sewbacca:
Rereading the code, I do not handle [`null` env unset requests](https://microsoft.github.io/debug-adapter-protocol/specification#Reverse_Requests_RunInTerminal). I suppose this would also affect integrated terminals, but it would need consideration of the consequences having `vim.NIL`s flying around:
https://github.com/mfussenegger/nvim-dap/blob/b8e54ac91cae94df6c609d986a7776e4ffc744ca/lua/dap/session.lua#L74-L76

mfussenegger:
> Rereading the code, I do not handle [`null` env unset requests](https://microsoft.github.io/debug-adapter-protocol/specification#Reverse_Requests_RunInTerminal). I suppose this would also affect integrated terminals, but it would need consideration of the consequences having `vim.NIL`s flying around:
> 
> https://github.com/mfussenegger/nvim-dap/blob/b8e54ac91cae94df6c609d986a7776e4ffc744ca/lua/dap/session.lua#L74-L76

I think that's okay for now. That can be fixed later if it ever causes an actual problem with an adapter making use of it.

PR #1323, opened 2024-08-26
Assign a safe default when the provider doesn't provide a column in set_cursor

GingerbreadFred:
Fixes [https://github.com/mfussenegger/nvim-dap/issues/1324](https://github.com/mfussenegger/nvim-dap/issues/1324)

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -372,6 +372,8 @@ end
 ---@param line integer
 ---@param column integer
 local function set_cursor(win, line, column)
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks for the PR but this seems to be a bug or regression in lldb-dap. See https://github.com/llvm/llvm-project/pull/73393

According to the specification the `column` property is required. I try to avoid merging workarounds for debug adapter bugs/spec violations.

PR #1338, opened 2024-09-25
suppport specify filetype when calling dap.continue

OrangeWatermelon:
I want to write a plugin similar to Idea, which can create configuration files for debugging. This way, debugging can be started from any location, rather than having to start debugging in the current buffer. While writing the script, I found that dap.continue can take opts as an argument, which led me to the select_config_and_run function. In this function, the filetype is obtained from the current buffer. However, further down, I noticed the line [opts.filetype = opts.filetype or filetype](https://github.com/mfussenegger/nvim-dap/blob/90616ae6ae40053103dc66872886fc26b94c70c8/lua/dap.lua#L503C5-L503C46), which seems to indicate that opts can pass in the filetype, but this filetype does not affect the configuration retrieval. Therefore, I modified the code so that the filetype is obtained from the buffer only when the filetype option is not passed in.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -11,7 +11,6 @@ local sessions = {}
 local session = nil
 local last_run = nil
 
-
 -- lazy import other modules to have a lower startup footprint
 local lazy = setmetatable({
   async = nil, --- @module "dap.async"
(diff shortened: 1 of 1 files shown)


mfussenegger:
Sorry but all the format changes make this rather hard to review and I'm not going to merge changes that have unrelated changes all over the place.

It also appears as if this would be a breaking change for the provider interface.

>  This way, debugging can be started from any location, rather than having to start debugging in the current buffer.

This is already the case if you're using either `.vscode/launch.json` files or a custom provider. See `:help dap-providers`.

PR #1349, opened 2024-10-11
Fix compare-with-nil error for resp.variablesReference

ttkingdom:
In case the `resp` variable has no `variablesReference` field, the comparison `resp.variablesReference > 0` will fail.

This commit add the is-not-nil check for `resp.variablesReference`.


mfussenegger:
See https://github.com/mfussenegger/nvim-dap/pull/1330#issuecomment-2336653220

Issue #1194, opened 2024-04-20
Make `require 'dap.ui.widgets'.hover()` jump to float only when calling twice


lawrence-laz:
### Problem Statement

Currently `require 'dap.ui.widgets'.hover()` jumps to the floating window on the first call, which requires to keep doing `<C-w>q` each time. Having to do this multiple times in short duration becomes very cumbersome (`dK` is mapped to dap hover):
```
dK<C-w>q ww dK<C-W>q ww dK<C-W>q ww  (and so on)
```

### Possible Solutions

Make it act like `vim.lsp.buf.hover()`:
```lua
--- Displays hover information about the symbol under the cursor in a floating
--- window. Calling the function twice will jump into the floating window.
function M.hover()
    -- ...
end
```
This would simplify the previous use case to:
```
dK ww dK ww dK
```

And for when you need to actually jump the floating window it's not hard to press the shortcut twice, as your fingers are already in place, as opposed to keep switching back an forth between `dK` and `<C-w>q`.

### Considered Alternatives

I tried looking for a way to configure this just for me, as I don't really mind the defaults being as they are now, but couldn't find how to do it.

mfussenegger:
Would using `dap.ui.widgets.preview()` work instead?
Or something like this:

```lua
local dap = require("dap")
vim.keymap.set("n", "<leader>di", function()
  dap.repl.open()
  dap.repl.execute(vim.fn.expand("<cexpr>"))
end)
vim.keymap.set("v", "<leader>di", function()
  -- getregion requires nvim 0.10
  local lines = vim.fn.getregion(vim.fn.getpos("."), vim.fn.getpos("v"))
  dap.repl.open()
  dap.repl.execute(table.concat(lines, "\n"))
end)
```


The focus-by-default for hover is going to stay, because it prioritizes the use-case where you need to expand some of the variables to drill down to the info you need. `preview` covers the "peek" use-case.

lawrence-laz:
These are ok, but not quite like built-in lsp hovering.

I think I managed to get what I wanted by modifying a copy of `dap.ui.widgets.hover()` to actually use a window created by `vim.lsp.util.open_floating_preview`:
```lua
function Dap_better_hover(expr, winopts)
    local value = dap_eval_expression(expr)

    local bufnr, winid = vim.lsp.util.open_floating_preview({}, "dap-float", {
        focusable = true,
        close_events = { 'CursorMoved', 'BufHidden', 'InsertCharPre' },
[...]

PR #1353, opened 2024-10-17
feat(widgets): quit floats with q

igorlfs:
Addresses a suggestion I made in https://github.com/mfussenegger/nvim-dap/issues/1194#issuecomment-2140501146.

### DIFF
--- a/lua/dap/ui/widgets.lua
+++ b/lua/dap/ui/widgets.lua
@@ -37,6 +37,7 @@ function M.new_cursor_anchored_float_win(buf)
     vim.wo[win].wrap = false
   end
   vim.bo[buf].filetype = "dap-float"
+  api.nvim_buf_set_keymap(buf, "n", "q", "<C-w>q", {})
   return win
 end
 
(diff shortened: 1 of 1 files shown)


mfussenegger (commented):
I'm not sure about this - if every plugin ends up mapping `q` then this should probably just be a default in neovim core instead for floating windows.

If it isn't a good default for all plugins, then I wonder why it should be done here. Having it work in one plugin but not another is odd.

If you do want to have it, you can register a FileType autocmd on `dap-float`, or you could even do a `WinEnter` and check if it's a floating window to set it generally.

igorlfs:
>  if every plugin ends up mapping `q` then this should probably just be a default in neovim core instead for floating windows.

That sounds like a better approach, but I'm unsure if other maintainers would be willing to accept such change (though it's common within core, e.g., for diagnostics and hover). If you're okay with this, I'll file an issue in core. 

> If it isn't a good default for all plugins, then I wonder why it should be done here. Having it work in one plugin but not another is odd.

IMO it's a good default for all plugins (if not, the vast majority). At least, most of the plugin I use, anyway.

> If you do want to have it, you can register a FileType autocmd on `dap-float`

That's what I currently have, but your second idea sounds better, thanks!

mfussenegger:
> If you're okay with this, I'll file an issue in core.

yes please.
I'm closing this here for now

igorlfs:
For reference, here's the issue on core: https://github.com/neovim/neovim/issues/31706

PR #1354, opened 2024-10-28
Add filter options to fix new unity debugger

blakekrpec:
As mentioned, [here](https://github.com/mfussenegger/nvim-dap/discussions/815), the new Unity debugger doesn't work without filterOptions passed for setBreakPointExceptions. This PR fixes that by including filterOptions. This enables use of the Unity debugger, and I believe shouldn't affect other adapters since filterOptions is always nil in session.lua.

Tested with Unity debugger, codelldb, and debugpy.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -147,9 +147,9 @@ M.defaults = setmetatable(
   {
     fallback = {
       exception_breakpoints = 'default',
-      ---@type string|nil
+      ---@type table[]|nil
       exception_options = nil,
-      ---@type string|nil
+      ---@type table[]|nil
       exception_filter_options = nil,
 
       ---@type "statement"|"line"|"instruction"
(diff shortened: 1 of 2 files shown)


mfussenegger:
`filterOptions` are optional per specification. This looks like it should be fixed in the unity debugger.

blakekrpec:
I agree. However, this Unity debugger comes from Microsoft, and is not open source. So the odds ever of getting that fixed seem very low to me. Thus, this PR. 

I understand not want to put band-aids in nvim-dap for a broken adapter, but I figured it was worth a shot since it seemed `filterOptions` was always `nil` in session.lua anyways.

blakekrpec:
I agree, the Unity debugger shouldn't be failing with an arg defined as optional in the spec. However, this debugger is from Microsoft, and I am not hopeful it will ever get fixed (thus this PR). But, I very much understand not wanting band-aids for a broken adapter in nvim-dap.

I feel there is still value in exposing the optional arg `filterOptions` in the same way that the optional arg `exceptionOptions` is currently exposed. Then I can at least set `filterOptions = {}` in my config to use nvim-dap with Unity debugger.

blakekrpec:
Closing this, and will open a new PR that allows users to pass optional args.

PR #1373, opened 2024-11-26
Fix errors on nil column send in the dap frame

adam-wolski:
It's possible for lldb dap to send nil as the column. Especially common with target.process.thread.step-in-avoid-nodebug 0 where after stepping finally finds a valid source first call is with a nil column leading to nvim lua errors due to arithmetic on nil.


mfussenegger:
Per spec the column is required, this needs fixing in lldb-dap: https://microsoft.github.io/debug-adapter-protocol/specification#Types_StackFrame

PR #1395, opened 2024-12-08
Allow terminal_win_cmd Callback to Receive Configuration.

nsoufian:
First, I want to express my gratitude for your incredible contributions to the Neovim ecosystem.
### Issue
In my current Neovim setup using `nvim-dap` for debugging, I utilize a custom console by defining a `terminal_win_cmd`.  However, with the existing implementation, there's a limitation: the callback function for `terminal_win_cmd` does not receive the debugging session's configuration. This makes it challenging to dynamically set the name of the integrated terminal based on the current debugging session, I need to set a name for the terminal on a multi-session scenario, enabling better picker search by name.

### Changes Made:
Modified the `terminal_win_cmd` callback to accept the `config` parameter. This allows the callback function to access details about the current debugging session, enabling dynamic naming of the integrated terminal.

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -563,6 +563,9 @@ The configuration values are set via `dap.defaults.fallback` (for global) or
   the integrated terminal. (See |dap-terminal|). Either a string or a function
   that must return a buffer number and a window number of the buffer/window
   for the terminal.
+  When using a function, it receives the current debug configuration as an 
+  argument, allowing for dynamic terminal naming and setup based on the 
+  session's configuration.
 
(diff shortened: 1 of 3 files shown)


nsoufian:
@mfussenegger Do you think this has a chance of being merged? In a multi-session scenario, I create an integrated terminal console using `terminal_win_cmd`, and I use a picker to switch between them. To enhance this functionality, I need the session's configuration context to be passed into `terminal_win_cmd`. This would allow me to set metadata for each terminal window, enabling more effective name-based searching within the picker.

mfussenegger:
Thanks - yes, makes sense to me - I might even make use of this myself

PR #1404, opened 2024-12-22
Add var_placeholders and var_placeholders_once to the 'dap' object

NoeMurr:
# Description
Make available var_placeholders to the user, so that they are able to modify them if required.

Also now ${workspaceFolderBasename} dependes on ${workspaceFolder} by default.

# Use case
${workspaceFolder} can be changed to for example the git root of the project.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -323,12 +323,12 @@ do
     return option
   end
 
-  local var_placeholders_once = {
+  M.var_placeholders_once = {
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks for the PR but I'm not going to expose this.

As a workaround you could wrap the existing `on_config` listener that expands the placeholders. See `:help dap-listeners-on_config`

There's also https://github.com/neovim/neovim/issues/8610 which if added might influence `${workspaceFolder}`.

PR #1405, opened 2024-12-23
Log stderr of adapters to dedicated files; improve spawn error messages

mfussenegger:


mthnglac:
Is this still an issue? Currently, I'm debugging ruby code with `ruby_lsp`, and the DAP console is still empty.
cc: @mfussenegger

Issue #1431, opened 2025-01-23
More customizable `dap.utils.pick_process`


NicholasMata:
### Problem Statement

I am currently using `dap.utils.pick_process` in a language specific extension I maintain. I have run into the following problems when trying to use solely `dap.utils`. 

#### Problem 1
Process names can sometimes get quite long which I believe was brought to attention by #1402 and then subsequently "resolved" by #1403. The resolution being to trim the process name.

#### Problem 2
The static title/prompt of "Select process: " might not be applicable once filter and custom label are applied.

### Possible Solutions

I would like to propose the following changes to make it a bit more flexible/customizable while maintaining existing functionality.

Add the following properties to `opts` in `dap.utils.pick_process`.

- label (optional): A function which takes in a proc object and returns a string (aka a label for that process)
- prompt (optional): A string which will be used as the prompt for pick window.

These changes will resolve both the "problems" outlined earlier by allowing the caller of the `dap.utils.pick_process` to customize the process label and pick prompt to their specific needs.

If you find the solution acceptable PR #1432 has them implemented.

### Considered Alternatives

Leave pick process prompt customization out of the scope of this library, and for each developer to create their own `pick_process` function if a custom process label or pick prompt is required.

---

Also just wanted to say thanks for the awesome plugin, makes debugging in Neovim a breeze and adding support for languages simple.

PR #1432, opened 2025-01-23
Customizable `dap.utils.pick_process` 

NicholasMata:
I have outline the proposal behind this PR in #1431. 

Brief recap is to make `dap.utils.pick_process` more customizable by allowing for custom process labels and custom prompt.

When refactoring I opted to leave `cols` and `wordlimit` variables from #1403 outside default label function as they don't need to be calculated for every process.


mfussenegger:
Makes sense to me. Thanks

 I added a fixup to turn the somewhat long inline type definitions into dedicated classes.

PR #1434, opened 2025-01-25
Use the correct caller fileinfo 

RoadRoller01:
Un the current implementation, it will only log the caller function. The _log caller function is the level wrapper function,
Therefor we set the level to 3 to get  the caller(level) caller(actual caller) function info.

More info here
https://www.lua.org/pil/23.1.html
and here 
https://www.lua.org/manual/5.4/manual.html#pdf-debug.getinfo

### DIFF
--- a/lua/dap/log.lua
+++ b/lua/dap/log.lua
@@ -86,7 +86,7 @@ function Log:_log(level, levelnr, ...)
   if argc == 0 then
     return true
   end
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks, I missed to update that in https://github.com/mfussenegger/nvim-dap/commit/3b796163591a300a433e25384f2574419e51da1a

PR #1436, opened 2025-01-26
Fix DapSetLogLevel when passing space after the level

RoadRoller01:
if you passed a space after the level it should just work you did not add any letters and its not a script.
In example if you typed:
“DapSetLogLevel DEBUG “
it will not work,
So I trimmed the function args, I also replaced `unpack(opts.fargs)` with `opts.args` because it will return the same result with less code 

Note:
`opts.fargs` is table of the function args.
`unpack(table)` convert table to string.
`opts.args` is args as string.
reference: https://neovim.io/doc/user/api.html#nvim_create_user_command()

### DIFF
--- a/plugin/dap.lua
+++ b/plugin/dap.lua
@@ -6,7 +6,7 @@ end
 local cmd = api.nvim_create_user_command
 cmd('DapSetLogLevel',
   function(opts)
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks

PR #1438, opened 2025-01-30
fix: set source filetype in providers.configs

catgoose:
Addresses issue where setting dap.defaults[ft]autostart = "config" uses "dap-repl" as filetype when attempting to start config from using repl and there is no active session


mfussenegger:
Thanks

PR #1448, opened 2025-02-21
Add overload type annotation for threads and stackFrame requests

igorlfs:
Hey!

Follow-up to https://github.com/mfussenegger/nvim-dap/commit/ffb077e65259f13be096ea6d603e3575a76b214a

I'm developing a plugin where I'm using these requests, so I thought it would be nice to have annotations for them.


mfussenegger:
Thanks

PR #1463, opened 2025-03-09
Add type annotations to initialize request

igorlfs:
Hello!

Another ant-sized PR, I hope you don't mind 😬 

[Spec](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Initialize)

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -90,7 +90,7 @@ M.repl = setmetatable({}, {
 ---@field exceptionInfo table<string, dap.RequestListener>
 ---@field goto table<string, dap.RequestListener>
 ---@field gotoTargets table<string, dap.RequestListener>
----@field initialize table<string, dap.RequestListener>
+---@field initialize table<string, dap.RequestListener<dap.Capabilities?, dap.InitializeRequestArguments>>
 ---@field launch table<string, dap.RequestListener>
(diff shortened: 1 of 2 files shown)


lua/dap.lua
```diff
@@ -90,7 +90,7 @@ M.repl = setmetatable({}, {
 ---@field exceptionInfo table<string, dap.RequestListener>
 ---@field goto table<string, dap.RequestListener>
 ---@field gotoTargets table<string, dap.RequestListener>
----@field initialize table<string, dap.RequestListener>
+---@field initialize table<string, dap.RequestListener<dap.InitializeResponse, dap.InitializeRequestArguments>>
```
mfussenegger, line 93:
```suggestion
---@field initialize table<string, dap.RequestListener<dap.Capabilities?, dap.InitializeRequestArguments>>
```


This might be a bit confusing but the listeners are called directly with the `body` of the response:

https://github.com/mfussenegger/nvim-dap/blob/8228cb00ab0850e483f8d58cc39cc580c05738d6/lua/dap/session.lua#L1080-L1082

This also means it doesn't need the `dap.InitializeResponse` definition

mfussenegger (commented):
> Another ant-sized PR, I hope you don't mind

I prefer smaller PRs, so that's good :+1:

PR #1480, opened 2025-03-30
fix: now pick_file can run on windows

jinzhongjia:
In the past, functions used Unix's check permissions to find executable files, which did not work on Windows.

Now change to use the suffix name and allow fallback to using powershell

### DIFF
--- a/lua/dap/utils.lua
+++ b/lua/dap/utils.lua
@@ -275,6 +275,9 @@ end
 ---@param opts {filter?: string|(fun(name: string):boolean), executables?: boolean}
 ---@return string[]
 local function get_files(path, opts)
+  local is_windows = vim.fn.has("win32") == 1 or vim.fn.has("win64") == 1
+  local uv = vim.uv or vim.loop
+  
   local filter = function(_) return true end
   if opts.filter then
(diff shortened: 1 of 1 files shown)


jinzhongjia:
Sorry, there is a problem, I need to make some changes

jinzhongjia:
I made a little formatting in the code, so there may be a little more places to modify commit. If you think my formatting style is not good, you can pull it into the local format and then push it up.

jinzhongjia:
And I add a little feat:
when `adapter.options.initialize_timeout_sec` is false, dap will not notify use adapter is slow, this is very useful on large projects
The company I work in is about 15 seconds to compile the go project. This slow prompt is sometimes a headache

mfussenegger:
Thanks. I cherry-picked the executable check change into https://codeberg.org/mfussenegger/nvim-dap/pulls/30 

For the timeout: You could set it to a higher value like 60 seconds instead. Not sure `false` support is warranted, the timer overhead should be negligible given that the rate of debug sessions created is usually low.

I left out the powershell fallback because `vim.fs.dir` has been available since neovim 0.8.0

Issue #1470, opened 2025-03-15
Feature Request: extend switchbuf to support "usevisibletab"


igorlfs:
### Problem Statement

Hello!

The "usevisible" option is very useful! I'm wondering if it can be further extended (or rather, a new option) to take into account other windows opened in the current tab, so that a jump only occurs if the line isn't visible on any window (in the current tab, obviously).

The main use case is that often, I'll create a split for a file, and in each window I'll focus on some part of the code. It's a workflow that gets disrupted if I "forget to jump" to the correct window after issuing a `:DapContinue`.

### Possible Solutions

Another option could be added to nvim-dap's extended switchbuf: the "usevisibletab" (the name could use some work), that would adhere to the behavior I described.

### Considered Alternatives

AFAIK there are no alternatives.

mfussenegger:
I'm not sure I understand the requirement, isn't that the same as `dap.defaults.fallback.switchbuf = "usevisible,usetab"` ?

igorlfs:
> isn't that the same as dap.defaults.fallback.switchbuf = "usevisible,usetab" ?

No, not really. The issue is that `"usevisible"` takes precedence over `"usetab"` when using this combination.

> I'm not sure I understand the requirement

Here's a demo, where I showcase the same file in 2 splits, each with different lines visible. When starting a session in the left split, it's used to focus the frame, in spite of the split on the right already showing the region where the session will be stopped. I want to avoid that: if any split (in the current tab) contains the region with the frame, I don't want to jump at all.

https://github.com/user-attachments/assets/13ca531a-707e-44df-93a5-645a1692fe97

I actually can't use 

```lua
require"dap".defaults.fallback.switchbuf = "usetab,usevisible"
```

either, since it doesn't take into account if the line is visible in the current window.

igorlfs:
Can be achieved with #1555

Issue #1478, opened 2025-03-28
jump to the frame use the main window


AllenFind:
### Problem Statement

My window layout is 

```
------------------------------
| frames |                   |
|        |                   |
---------  CODE WINDOW       |       
| repl   |                   |
|        |                   |
---------                    |
| scopes |                   |
|        |                   |
------------------------------
```

I know we have switchbuf to choose useopen/uselast...

I wonder if we can always use the code window when I jump to the different frame. 'uselast' is helpful but it requires the previous window is code window. There was a case when I first type some commands in the 'repl' window then use the frames.

Any suggestions?

mfussenegger:
One problem here is how to identify the code window. Naive option is to look for a buffer that's operating on a file, but that breaks stepping in synthethic sources. So it must be able to jump to a window with any buffer, which unfortunately includes windows like the frames or scopes from other plugins.

But with https://github.com/mfussenegger/nvim-dap/pull/1555 you can customize the switchbuf behavior with a function to have complete control over it and define your own heuristics.

PR #1485, opened 2025-04-09
Add tabusevisible switchbuf option

igorlfs:
Addresses #1470 

I've been using this for a couple of weeks without any issues. But I've been holding back onto opening the PR because I don't think that's a long-term solution: adding variations to the `switchbuf` option is awkward. 

You'd think that this would be the last one, but we can go further. For instance, the `usetab` value does not take into account if there's a window where the line is visible (akin to the `usevisible` option) within the same tab. Instead, with `usetab`, the first window that matches the buffer is always used, regardless of the line being visible elsewhere. 

Thoughts?

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -592,11 +592,12 @@ The configuration values are set via `dap.defaults.fallback` (for global) or
                        See |'switchbuf'|. Defaults to the global `'switchbuf'`
                        setting.
 
-                       nvim-dap provides an additional `usevisible` option
-                       that can be used to prevent jumps within the active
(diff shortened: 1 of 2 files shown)


mfussenegger:
I'm also not sure where to draw the line here - currently I'm tending more towards simply allowing `switchbuf` to be a function and let users customize it as they see fit.


Also given issues like https://github.com/mfussenegger/nvim-dap/issues/1478
Accounting for all the different buftypes/layouts/behaviors that people might want seems to be a never ending story otherwise.

Issue #1488, opened 2025-04-15
Interactive terminal programs display incorrectly in the nvim-dap provided terminal


miroshQa:
### Debug adapter definition and debug configuration

```lua
local dap = require("dap")

-- https://github.com/mfussenegger/nvim-dap/wiki/C-C---Rust-(gdb-via--vscode-cpptools)
dap.adapters.cppdbg = {
  id = 'cppdbg',
  type = 'executable',
  command = '/absolute/path/to/cpptools/extension/debugAdapters/bin/OpenDebugAD7',
}

dap.configurations.c = {
  {
    name = "Launch file (cpp tools)",
    type = "cppdbg",
    request = "launch",
    program = "${command:pickFile}",
    cwd = '${workspaceFolder}',
    stopAtEntry = true,
  },
}
```

### Debug adapter version

main branch

### Steps to Reproduce

1. Clone the neovim repository `git clone https://github.com/neovim/neovim.git`
2. Compile with debug symbols `make CMAKE_BUILD_TYPE=Debug CMAKE_C_COMPILER=clang`
3. Open the main.c file `nvim src/nvim/main.c`
4. Launch the program (./build/bin/nvim)
5. Press continue so neovim appears in the terminal

### Expected Result

Neovim displays correctly (and other advanced TUI programs). 
### Actual Result

First, I show it working fine in my default terminal, then demonstrate how it looks in the DAP terminal.
As you can see, Neovim displays incorrectly (and the same goes for other interactive programs like this).

https://github.com/user-attachments/assets/36ea4158-df0d-4a8f-8adb-ec4b351a178a

I suspected this was a debug adapter issue, but even with lldb-dap, the problem persists. Also, I don’t have this problem in vs code, and if we launch neovim in the neovim builtin terminal (:terminal), then everything also works fine. So I suppose it’s some problem with the way nvim-dap handles the terminal.

miroshQa:
I just figured out that if I don't use any UI plugins and just use the terminal window initially created by nvim-dap, everything works well.

![Image](https://github.com/user-attachments/assets/7811a1c7-1dd0-4806-9621-1ea1ba7b9dff)

I guess this means the nvim-dap terminal should handle terminal resizing. Or does it already handle resizing, but only for the initially created window—meaning if we open this terminal buffer in another window, resizing isn't handled?

mfussenegger:
[...]

PR #1490, opened 2025-04-18
replace nvim_open_term with vim.jn.jobstart(..., {term = true})

miroshQa:
This PR is a continuation of #1488, where I explored some nvim-dap terminal issues—specifically, wrong scrollback handling and the cursor not changing its shape when debugging Neovim. Resizing problems, as I found out, were due to the debug adapter, so this PR doesn’t fix those.

While continuing to use it, I faced other issues with this terminal. I launched tests in my project, and for some reason, the test output didn’t have colors. With this patch, all three issues I found disappeared. It also remove the fragile nvim_chan_send logic and lets neovim handle it itself, which I think is always for the best. The 'Process exited' and 'CR' used to delete the buffer at the end are also removed because neovim :terminal already handles them

So this patch makes the nvim-dap-provided terminal behave exactly like the terminal you get using :terminal cmd, without the strange quirks we get when creating a terminal using nvim_open_term.

I also saw your comments about using termopen before and its problems, but with this implementation, it doesn’t have such drawbacks (if I understand those issues correctly):
https://github.com/mfussenegger/nvim-dap/issues/1488#issuecomment-2807530368

In the video, I show the third issue I found with the terminal created by nvim_open_term. First, I show how it works now: the output doesn’t have colors (in this case, for some reason). Then I apply this patch and show how it works after this PR. Now you can see everything works as expected. The output has colors.

https://github.com/user-attachments/assets/5873c297-2929-44c8-bf5c-eedccc592cd7

If this is still not convincing enough, I can provide additional videos showcasing issues with scrollback.

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -262,42 +262,21 @@ local function run_in_terminal(lsession, request)
   if path and path ~= "" then
     vim.bo[terminal_buf].path = path
   end
-  local jobid
 
-  local chan = api.nvim_open_term(terminal_buf, {
-    on_input = function(_, _, _, data)
-      pcall(api.nvim_chan_send, jobid, data)
-    end,
-  })
-  local opts = {
-    env = next(body.env or {}) and body.env or vim.empty_dict(),
-    cwd = (body.cwd and body.cwd ~= '') and body.cwd or nil,
-    height = terminal_win and api.nvim_win_get_height(terminal_win) or math.ceil(vim.o.lines / 2),
-    width = terminal_win and api.nvim_win_get_width(terminal_win) or vim.o.columns,
-    pty = true,
-    on_stdout = function(_, data)
-      local count = #data
-      for idx, line in pairs(data) do
-        if idx == count then
-          local send_ok = pcall(api.nvim_chan_send, chan, line)
-          if not send_ok then
-            return
-          end
-        else
-          local send_ok = pcall(api.nvim_chan_send, chan, line .. '\n')
-          if not send_ok then
-            return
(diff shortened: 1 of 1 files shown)


miroshQa:
Good news for me: It actually does fix the resize issue when using lldb-dap to debug Neovim.

It still breaks with cpptools, but that is definitely not nvim-dap's fault.

lua/dap/session.lua
```diff
@@ -262,42 +262,21 @@ local function run_in_terminal(lsession, request)
...
-          end
-        end
+  local jobid
+  vim.api.nvim_buf_call(terminal_buf, function()
+    jobid = vim.fn.jobstart(body.args, {
+      env = next(body.env or {}) and body.env or vim.empty_dict(),
+      cwd = (body.cwd and body.cwd ~= '') and body.cwd or nil,
+      height = terminal_win and api.nvim_win_get_height(terminal_win) or math.ceil(vim.o.lines / 2),
+      width = terminal_win and api.nvim_win_get_width(terminal_win) or vim.o.columns,
+      term = true,
```
mfussenegger, line 273: `jobstart` with `term = true` is new in 0.11, for 0.10 this will need to use `termopen` or stick to the old logic.

mfussenegger (commented):
On a quick test this seems to work fine. @rcarriga can you confirm that this doesn't break anything for nvim-dap-ui?

rcarriga:
Thanks for the ping, looks all good from my end :+1:

mfussenegger:
Thanks

PR #1495, opened 2025-04-25
Fix terminal buffer name incorrectly overwriten by vim.fn.jobstart

wenjinnn:
Hello.
After #1490 was merged, the terminal buffer name becomes the command name instead of the  `[dap-terminal] ...` format, like this.

I find out that the terminal buffer name was being incorrectly overwritten after `vim.fn.jobstart` called.
Fixed by setting terminal buffer name after `vim.fn.jobstart`.


mfussenegger:
Thanks

PR #1498, opened 2025-05-01
REPL commands for stepping with instruction granularity

Jorenar:
Hi!

As in the title, this PR adds REPL commands for stepping with instruction granularity regardless of what is set as the default stepping granularity.

It also fixes name of `granularity` parameter in docs.


Jorenar:
@mfussenegger ?

Jorenar:
Yeah, GDB simply adds "i" suffix to stepping commands, and I've did the same here as it's quite convenient.

I'm playing around with making a plugin adding disassembly view ([nvim-dap-disasm](https://github.com/Jorenar/nvim-dap-disasm)), so having instruction stepping commands is helpful.

Jorenar:
@mfussenegger Could we merge?

Issue #1037, opened 2023-09-11
REPL autocompletion is a bit messy with vscode-cpptools.


Alexey104:
### Debug adapter definition and debug configuration

```
local home=os.getenv( "HOME" )
local gdb_path = home .. "/.config/nvim/gdb/extension/debugAdapters/bin/OpenDebugAD7"

local dap = require('dap')

dap.adapters.cppdbg = {
  id = 'cppdbg',
  type = 'executable',
  command = gdb_path,
}

dap.configurations.cpp = {
  {
    name = "Launch file",
    type = "cppdbg",
    request = "launch",
    program = function()
      return vim.fn.input('Path to executable: ', vim.fn.getcwd() .. '/', 'file')
    end,
    cwd = '${workspaceFolder}',
    stopAtEntry = true,
  },
  {
    name = 'Attach to gdbserver :1234',
    type = 'cppdbg',
    request = 'launch',
    MIMode = 'gdb',
    miDebuggerServerAddress = 'localhost:1234',
    miDebuggerPath = '/usr/bin/gdb',
    cwd = '${workspaceFolder}',
    program = function()
      return vim.fn.input('Path to executable: ', vim.fn.getcwd() .. '/', 'file')
    end,
  },
}

setupCommands = {
  {
     text = '-enable-pretty-printing',
     description =  'enable pretty printing',
     ignoreFailures = false
  },
}
```

### Debug adapter version

vscode-cpptools v1.17.5

### Steps to Reproduce

1) Create a functions called `some_function()`.
2) Open the REPL window, switch to it and enter insert mode.
3) Type `-exec b some`:
![repl](https://github.com/mfussenegger/nvim-dap/assets/38672094/536700cf-80f0-4048-8e6c-4916ea13aa99)
4) Press `<C-x><C-o>` to trigger autocompletion:
![completion](https://github.com/mfussenegger/nvim-dap/assets/38672094/048b1f43-a3e1-4c5f-8981-ea014438fe91)

### Expected Result

The completed suggestion should overwrite the entire line:
`-exec b some_function()`

### Actual Result

The completed suggestion is appended to what was typed before triggering the completion:
`-exec b -exec b some_function()`

mfussenegger:
Can you try with https://github.com/mfussenegger/nvim-dap/pull/1038 ?  It should fix this.

Alexey104:
Confirm, this fix works.
Thank you so much!

PR #1507, opened 2025-05-15
More handling of length/start in completions response in repl.omnifunc

Jorenar:
Continuation of #1037 and #1038
Bit of a workaround for microsoft/debug-adapter-protocol#138

---

To add to this confusing mess, some servers do send whole line back, some do it without `start` parameter, some do it with `start = 0`. This PR provides workaround for these situations.


mfussenegger:
Can you post some example payloads and a description of before/after behaviour or even better add them as test cases?

Jorenar:
Unfortunately, tests didn't want to cooperate with me, and I don't have enough Lua experience to write them by trial and error.

But I've updated the PR description with examples of the problems.

mfussenegger:
For the second case the behavior without your change actually seems more correct to me. Given that the debug adapter is reporting the `start` property - I think the client should honor that as word boundary instead of inferring it based on prefix matching.

How would that otherwise behave in cases where there is a mix? - If the adapter provides a result with some prefix matches and some without because it is doing fuzzy matching.

mfussenegger:
> The flag prefixed would be set to false, since not all results are with same prefix

I meant more in regards to user facing behavior - I have the impression that you'd get inconsistency from a single adapter, where sometimes completion refers to start of line, and sometimes it doesn't.

Now it might be inconsistent between different adapters, but at least for the same adapter it should be consistent - assuming the adapter itself is.

Jorenar:
Hmm, I understand.
In that case, I'll remove prefix trimming, and leave only the fix for determining `start`

mfussenegger:
> In that case, I'll remove prefix trimming, and leave only the fix for determining start

Thanks - I made a couple more tweaks; if I didn't mess up the tests it should still cover the gdb case.

PR #1519, opened 2025-06-01
Call handlers on session change and ability to disable terminal buffer pooling

Run1e:
Hi there!

Me and @igorlfs are working on a [PR to nvim-dap-view](https://github.com/igorlfs/nvim-dap-view/pull/51) which will enable multi-session support.

There are two problems that prevent us from implementing this feature fully, namely:
1. It's not possible for `nvim-dap-view` to know when the dap session changes (so we can't show the active sessions terminal buffer when it changes).
2. `nvim-dap` will pool terminal buffers, which makes it hard/impossible for us to know which one is being used for which session.

I've tried to implement some minimal changes that fixes these issues for us:
1. Instead of opaquely setting the `session` in various places, call [`set_session`](https://github.com/Run1e/nvim-dap/commit/87ad93697c16d208630bf5ca0d8d5ea29ad4fb40#diff-b59b772740a7d213d40288d3c87a48040714405a93860817efa817d8cc2e2a94R558-R564) which also calls a list of handlers that other plugins can register using [`register_on_session_changed_handler`](https://github.com/Run1e/nvim-dap/commit/87ad93697c16d208630bf5ca0d8d5ea29ad4fb40#diff-b59b772740a7d213d40288d3c87a48040714405a93860817efa817d8cc2e2a94R552-R556).
2. Implement two functions that can be used to [enable or disable](https://github.com/Run1e/nvim-dap/commit/87ad93697c16d208630bf5ca0d8d5ea29ad4fb40#diff-1c238a5ffb6d36f9412528c8783aacf498b7c4456f7ce2131779e1ea04bb98b2R174-R182) the terminal buffer pooling.

Disabling the terminal buffer pooling has the effect that each new session will *always* call the `terminal_win_cmd` function which in practice means plugins like `nvim-dap-view` can gain control over how terminal buffers are created and used for sessions if they need to.

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -171,6 +171,15 @@ local function create_terminal_buf(terminal_win_cmd, config)
   end
 end
 
+local enable_termbuf_pooling = true
+
+Session.termbuf_pooling_disable = function()
+  enable_termbuf_pooling = false
+end
+
+Session.termbuf_pooling_enable = function()
+  enable_termbuf_pooling = true
+end
 
 local terminals = {}
 do
@@ -180,12 +189,14 @@ do
   ---@param filetype string
   ---@return integer, integer|nil
   function terminals.acquire(win_cmd, config, filetype)
-    local buf = next(pool)
-    if buf then
-      pool[buf] = nil
-      if api.nvim_buf_is_valid(buf) then
-        vim.bo[buf].modified = false
-        return buf
+    if enable_termbuf_pooling then
+      local buf = next(pool)
+      if buf then
+        pool[buf] = nil
+        if api.nvim_buf_is_valid(buf) then
+          vim.bo[buf].modified = false
+          return buf
+        end
       end
     end
     local terminal_win
@@ -258,6 +269,7 @@ local function run_in_terminal(lsession, request)
   end
 
   local jobid
+  local on_exit = enable_termbuf_pooling and function() terminals.release(terminal_buf) end or nil
   vim.api.nvim_buf_call(terminal_buf, function()
     local termopen = vim.fn.has("nvim-0.11") == 1 and vim.fn.jobstart or vim.fn.termopen
     jobid = termopen(body.args, {
@@ -266,9 +278,7 @@ local function run_in_terminal(lsession, request)
       height = terminal_win and api.nvim_win_get_height(terminal_win) or math.ceil(vim.o.lines / 2),
       width = terminal_win and api.nvim_win_get_width(terminal_win) or vim.o.columns,
       term = vim.fn.has("nvim-0.11") == 1 and true or nil,
-      on_exit = function()
-        terminals.release(terminal_buf)
-      end
+      on_exit = on_exit,
     })
   end)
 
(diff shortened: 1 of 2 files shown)


Run1e:
Slightly more in-depth explanation of the terminal buffer pooling issue can be found in this comment on the PR: https://github.com/igorlfs/nvim-dap-view/pull/51#issuecomment-2849213196

MironPascalCaseFan:
My 2 cents:

1 - I think this session event should go in dap.listeners.on_session_changed (see :h dap-listeners-ext)

2 - My advocacy for the ability to disable terminal pooling:
I assumed that terminal_win_cmd is called each time the debug adapter requests a terminal, so I can provide my own buffer and my plugin has full control over it. But because of this pooling behavior, that’s no longer guaranteed. If the buffer wasn’t removed, it will reuse it and won’t call terminal_win_cmd anymore.

I understand it's useful for nvim-dap internally so it doesn’t uncontrollably create buffers for each new session and "leak" them, but some plugins would ideally want to control terminals buffer management themselves, so the option to disable pooling is necessary.

Run1e:
> I think this session event should go in dap.listeners.on_session_changed (see :h dap-listeners-ext)

Thanks! Not entirely sure how I missed that there were client events I could just add this to. I've added a `on_session_change` event.

lua/dap/session.lua
```diff
@@ -171,6 +171,15 @@ local function create_terminal_buf(terminal_win_cmd, config)
...
 
+local enable_termbuf_pooling = true
+
+Session.termbuf_pooling_disable = function()
+  enable_termbuf_pooling = false
+end
+
+Session.termbuf_pooling_enable = function()
+  enable_termbuf_pooling = true
+end
```
miroshQa, line 182: I think it can be just session property like Session.terminal_pooling_enabled = boolean

igorlfs:
> It's not possible to know when the dap session changes

Is this abstraction really necessary? We can create the listeners for the corresponding events. 

Also, shouldn't session termination events also fire `on_session_change`?

Run1e:
> > It's not possible to know when the dap session changes
> 
> Is this abstraction really necessary? We can create the listeners for the corresponding events.
> 
> Also, shouldn't session termination events also fire `on_session_change`?

You and the maintainers of this repo would probably be the ones to answer that.

I went through and looked at everywhere the `session` variable gets set to a new value, and everywhere that happens a `on_session_change` event should now fire. Assuming I didn't miss a spot, and terminating a session sets a new session or sets `session = nil`, the event should fire. Please let me know if things aren't working the way you'd want them to!

mfussenegger:
I've extracted the first part of this with a few changes to https://codeberg.org/mfussenegger/nvim-dap/pulls/1.
Please check if the behavior is sufficient to cover the use-case.

I'll address the terminal buffer pooling separately.

Run1e:
> Please check if the behavior is sufficient to cover the use-case.

Works just fine with my experimental implementation, so I'd imagine it would work for @igorlfs's planned implementation too.

igorlfs:
> > Please check if the behavior is sufficient to cover the use-case.
> 
> Works just fine with my experimental implementation, so I'd imagine it would work for @igorlfs's planned implementation too.

Indeed, everything is also working fine here 👍🏼

mfussenegger:
With https://github.com/mfussenegger/nvim-dap/pull/1546 and https://codeberg.org/mfussenegger/nvim-dap/pulls/1 am I correct that the use-cases are now covered and this can be closed?

igorlfs:
> am I correct that the use-cases are now covered and this can be closed?

Yep

PR #1520, opened 2025-06-06
Extract the logic of awaiting a bunch of async function results into a separate function, await_all. fix #1491

### LINKED ISSUE #1491: The scopes widget doesn't display global variables if there are no local variables

Not related to the debug adapter—I also reproduced it while debugging C/C++ code with lldb-dap. You can reproduce it with any adapter when you enter a function that doesn’t have local variables. But here’s the debugpy configuration reference: https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation#python

### Debug adapter version

main branch

### Steps to Reproduce

0. Setup commands to open and refresh scopes widget
```lua
vim.api.nvim_create_user_command("OpenScopes", function()
  local widgets = require("dap.ui.widgets")
  local scopes = widgets.sidebar(widgets.scopes)
  scopes.open()

  vim.api.nvim_create_user_command("RefreshScopes", function()
    scopes.refresh()
  end, {})
end, {})
```

1. Make an new directory and create the main.py file there 
```py
def no_local_variables():
    print("foo")
    print("boo")
    print("bar")


a = 2
b = 3
no_local_variables()
```
2. Set a breakpoint on the line with variable `a` and launch debugging (debugpy adapter).
3. Open the scopes widget using the "OpenScopes" command
4. Step over until the no_local_variables function call and then do step into.
5. You will see that the scopes widget doesn't show anything. Then manually refresh it using RefreshScopes
6. Now the scopes widget shows global variables again. Make a step over and you'll see that the scopes widget doesn't show anything again

### Expected Result

The scopes widget refreshes correctly, and we can see all global variables without manually refreshing it after each step-over when we are inside a function without local variables.

### Actual Result

https://github.com/user-attachments/assets/794d2366-a985-4291-845a-baeab6302b20

miroshQa:
1. Extract the logic of awaiting a bunch of async function results into a separate util function, await_all.

2. Fixes the bug that caused the scopes widget to turn blank when trying to expand an empty collection. That was happening because the callback wasn’t called if the unloaded (#variables) was equal to 0.
From my tests, it also fixes the issue #1491.

### DIFF
--- a/lua/dap/async.lua
+++ b/lua/dap/async.lua
@@ -1,9 +1,9 @@
-local M = {}
+local async = {}
 
 --- Run a function in a coroutine with error handling via vim.notify
 ---
 --- If run is called within a coroutine, no new coroutine is created.
-function M.run(fn)
+function async.run(fn)
   local co, is_main = coroutine.running()
   if co and not is_main then
     fn()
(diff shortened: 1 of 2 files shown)


MironPascalCaseFan:
My main goal was to fix #1491, so after investigation, I figured out that it happens because the callback isn’t called if the unloaded equals zero. That could be fixed just by adding an additional if-check, but I decided to extract this await logic into a separate function because it’s a pretty common thing to do and could be useful in the future (if you try to implement recursive expand, for example) If you don’t need this, I understand—just add a check: if `unloaded == 0`, then `cb(variables)` to fix #1491

mfussenegger:
Thanks, this was fixed with https://codeberg.org/mfussenegger/nvim-dap/commit/2edd6375692d9ac1053d50acfe415c1eb2ba92d0

PR #1538, opened 2025-07-20
Fix first line in doc for local-additions

jpc-hpe:
The documentation for vim help pages states that a first line is required for the file to appear under :help local-additions, with the following format: *tag* description

This change modifies the first line to follow that format


mfussenegger:
Thanks

PR #1542, opened 2025-08-06
feat(widgets): add 'q' mapping to close centered float widget for convenience

abishz17:
This PR adds buffer-local 'q' mapping in `new_centered_float_win(buf)`  widget to close floating windows.

**Motivation:**
Users currently need to type `:q<CR>` to close the dap floating window, which is less convenient than the standard `q` key.

### DIFF
--- a/lua/dap/ui/widgets.lua
+++ b/lua/dap/ui/widgets.lua
@@ -65,6 +65,7 @@ function M.new_centered_float_win(buf)
     vim.wo[win].wrap = false
   end
   vim.bo[buf].filetype = "dap-float"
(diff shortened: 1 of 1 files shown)


igorlfs:
See https://github.com/mfussenegger/nvim-dap/pull/1353

abishz17:
Makes sense tbh. I wish neovim at its core already made 'q' mapping the default widget closing binding.

igorlfs:
>  I wish neovim at its core already made 'q' mapping the default widget closing binding.

There's an ongoing discussion [here](https://github.com/neovim/neovim/issues/31706) (although stalled). For the time being, you can create an autocmd to close floats.

PR #1543, opened 2025-08-16
fix: sign id is not globally unique

tomtomjhj:
Problem:
The sign id returned by sign_place() is buffer-local, so the ids of
signs in different buffers can be the same. Due to this, set_breakpoint
with condition overrides condition of another breakpoint in another
buffer.

Solution:
* Maintain mapping of buffer → sign id → breakpoint.
* Also fix other issues with wrong fields of bp.


tomtomjhj:
lint failure seems unrelated to this pr

mfussenegger:
Thanks, nice catch. A bit surprised that I haven't noticed this for so long.

PR #1546, opened 2025-08-26
Retrieve `term_buf` by session

igorlfs:
Implements https://github.com/mfussenegger/nvim-dap/discussions/1523#discussioncomment-14215672

Alternative to #1519 

**Problem**: UI plugins need to accurately retrieve the terminal buffer for the current session, so they can show the corresponding buf.

**Solution**: add an optional field to `dap.Session` that stores the buffer.

By the way, I'm planning on making some more contributions, would you prefer if new PRs were opened on codeberg?

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -56,6 +56,7 @@ end
 ---@field private handle uv.uv_stream_t
 ---@field current_frame dap.StackFrame|nil
 ---@field initialized boolean
+---@field term_buf? integer
 ---@field stopped_thread_id number|nil
 ---@field id number
 ---@field threads table<number, dap.Thread>
@@ -258,6 +259,7 @@ local function run_in_terminal(lsession, request)
   end
 
   local jobid
+  lsession.term_buf = terminal_buf
   vim.api.nvim_buf_call(terminal_buf, function()
     local termopen = vim.fn.has("nvim-0.11") == 1 and vim.fn.jobstart or vim.fn.termopen
     jobid = termopen(body.args, {


mfussenegger:
> Problem: UI plugins need to override terminal_win_cmd to prevent nvim-dap from auto-spawning the terminal with belowright new. By doing that, they also inherit the responsibility of maintaining the state of buffers, which may not play very well with terminal buf pooling.

> Solution: Allow overriding terminal_win_cmd without taking ownership of buffers by setting it to nil.

I'm a bit confused by this, what's the difference between `terminal_win_cmd = nil` with this PR and the following without it?

```lua
terminal_win_cmd = function()
    return api.nvim_create_buf(false, false)
end
```

Also - one disadvantage of not creating the window is that the terminal won't have the correct `height`/`width` information, which means word wrapping can be messed up in the terminal output.


---


The `term_buf` addition looks okay to me.


---

> By the way, I'm planning on making some more contributions, would you prefer if new PRs were opened on codeberg?

Yes please. Also if it's bigger changes upfront issue/discussion would be appreciated.

igorlfs:
> I'm a bit confused by this, what's the difference between terminal_win_cmd = nil with this PR and the following without it?

I'm sorry, you're right. I assumed that if the buffer was created externally, its lifecycle should also be handled externally, which is just not the case. I'll drop the commit.

> Also - one disadvantage of not creating the window is that the terminal won't have the correct height/width information, which means word wrapping can be messed up in the terminal output.

IMO, it's not worth it worrying about that. It's a common workflow, and I'm not aware of any related reports -- at least not on nvim-dap-view's side.

igorlfs:
I'm playing around a little more and I think the current implementation is incomplete. I'm testing with the `js-debug-adapter` and its nested sessions.

Currently, the top level session gets assigned the terminal buffer and not its children, which is "incorrect". Any ideas on how to handle this situation?

EDIT, to give more context: I'm adding a feature to navigate between sessions. When I run a configuration for the `js-debug-adapter`, 2 sessions are created:

```
→ 1: Launch JS                 -- created via `runInTerminal`
    2: index.js [1208942]      -- created via `startDebugging`
```

When a breakpoint is hit, session 2 is focused. It's the session that's actually controlling the terminal, in spite of not "owning" it. If I switch to session 1 (i.e., the one that thinks it owns the terminal), I can't control anything about the session (e.g, I can't step) because session 1 is _not_ stopped.

Within nvim-dap-view, if I then switch back to session 2 and visit the console, it won't show anything, because session 2 _owns no terminal_.

Ideally, the child would "own" the terminal. To me, that's a more accurate representation of who's actually controlling the terminal. But I have no idea how that would work with multiple children, or deeper nesting. Thoughts?

mfussenegger:
> Ideally, the child would "own" the terminal. To me, that's a more accurate representation of who's actually controlling the terminal. But I have no idea how that would work with multiple children, or deeper nesting. Thoughts?

Could you traverse the hierarchy via `session.parent` to find the root and its `term_buf`?
Technically the existing representation is accurate given that the root session sends `runInTerminal`. Afaik currently no debug adapter sends runInTerminal from a child session - but so far debugpy & vscode-js-debug are also the only adapters that use the `startDebugging` functionality.


> If I switch to session 1 (i.e., the one that thinks it owns the terminal), I can't control anything about the session (e.g, I can't step) because session 1 is not stopped.

If it helps maybe the `M.set_session(first_stopped_session())` in the step/continue functions could be extended so that `first_stopped_session()` also traverses into children. But that would also mean that it would switch back automatically from 1 to 2.

igorlfs:
> Could you traverse the hierarchy via `session.parent` to find the root and its `term_buf`?

Yes, that's an alternative I considered, but ideally I'd like to avoid having to "special case" the `js-debug-adapter`. 

> Technically the existing representation is accurate given that the root session sends runInTerminal

On second thought, I agree with that. It's just a slightly unfortunate situation.

> but so far debugpy & vscode-js-debug are also the only adapters that use the startDebugging functionality

Could you share a minimal setup that I can use to ensure my implementation works with debugpy?

> If it helps maybe the M.set_session(first_stopped_session()) in the step/continue functions could be extended so that first_stopped_session() also traverses into children.

I'm afraid this just complicates the logic for stepping, and I fear it may introduce unexpected changes elsewhere. Not sure if these concerns are justified, though.

---

For the `js-debug-adapter`, my plan is to traverse the parents and fetch the root's term_buf. That's for the children, as I'll handle the root by just skipping it.

That being said, I think this PR will not require any further changes, and it could be merged as is.

PR #1552, opened 2025-09-10
check for windows for breakpoint request

DePaWSiT:
replace breakpoint path for windows from / to \\

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -943,6 +943,10 @@ do
         api.nvim_buf_attach(bufnr, false, { on_detach = remove_breakpoints })
       end
       local path = api.nvim_buf_get_name(bufnr)
+      local is_windows = vim.fn.has('win32') == 1
+      if is_windows then
+        path = path:gsub("/", "\\")
+      end
       ---@type dap.SetBreakpointsArguments
       local payload = {
         source = {


DePaWSiT:
This was only tested for C# using netcoredbg so I don't know if it will cause problems for other DAPs on windows

mfussenegger:
Thanks for the PR.

Created https://github.com/microsoft/debug-adapter-protocol/issues/563 to see how to deal with this properly. As it is I think it could break remote debugging scenarios and nvim-dap core must also stay debug adapter agnostic. There must not be any checks like ` self.config.type == "coreclr"`

DePaWSiT:
Hello,

The problem has kinda been solved (in my instance at least), I combine nvim-dap with easy-dotnet for debugging using netcorebdg. Somebody else mentioned the same problem (the mention above your message) which was solved by a small update which pretty much does what this pull request does (the problem might appear open but update at the bottom of the thread solved the issue).

It was mentioned in the other thread that snacks.picker could overwrite the shellslash option but even if that were the case, the problem will still occur, it will just make it under very specific circumstance.

Anyway the problem is solved for me (for now), do i close this or leave it open?

mfussenegger:
> Anyway the problem is solved for me (for now), do i close this or leave it open?

I'll close it for now and revisit once https://github.com/microsoft/debug-adapter-protocol/issues/563 gets a reply

PR #1557, opened 2025-09-27
Set cursor outside custom switchbuf function

tomtomjhj:
Setting the cursor is common to any switchbuf function and should handle error. So nvim-dap should set the cursor on behalf of users.

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -598,8 +598,9 @@ The configuration values are set via `dap.defaults.fallback` (for global) or
   example: 'usevisible,usetab,uselast'
 
   For more complex use cases, nvim-dap allows overriding with a function. The
-  function receives 3 arguments: (bufnr, line, column). It has full control of
-  how Neovim should behave and it is not expected to return anything.
+  function receives bufnr and should return a window id with its current
+  buffer set to the given buffer. It has full control of how Neovim should
+  behave.
 
 
 - `on_output`. A function with two parameters: `session` and `output_event`:
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -409,7 +409,7 @@ end
 ---@param bufnr number
 ---@param line number
 ---@param column number
----@param switchbuf string|fun(bufnr: integer, line: integer, column: integer):nil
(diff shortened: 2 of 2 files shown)


mfussenegger (commented):
This would be a breaking change in the way it is and take away the flexibility to suppress cursor movements.

I think a compromise to make this non-breaking and support both use cases would be to keep the 3 parameters and make the return value optional.

If it returns a `win` number it will set the cursor, if it returns `nil` it wont.

cc @igorlfs

igorlfs:
I don't know if my use case is "over-engineered", but my implementation usually _does not_ set the cursor position. Here's the [full implementation](https://gist.github.com/igorlfs/de828149d41196eea71c23e7997130d9).

> I think a compromise to make this non-breaking and support both use cases would be to keep the 3 parameters and make the return value optional.

This would work for me. Keeping the line parameter is a must: it can be used to know when _not to set the cursor position_ (in a way similar to `usevisible`). No strong feeling about the column parameter, though.

doc/dap.txt
```diff
@@ -598,8 +598,9 @@ The configuration values are set via `dap.defaults.fallback` (for global) or
   example: 'usevisible,usetab,uselast'
 
   For more complex use cases, nvim-dap allows overriding with a function. The
-  function receives 3 arguments: (bufnr, line, column). It has full control of
-  how Neovim should behave and it is not expected to return anything.
+  function receives bufnr and should return a window id with its current
+  buffer set to the given buffer. It has full control of how Neovim should
```
igorlfs, line 602: Side note: I added this comment about "full control" _because_ there was no return value. IMO, it feels a little redundant when the function is expected to return a window number.

tomtomjhj:
oh ok I wasn't aware of such a use case. I think supporting both signatures is a bit too complex.

igorlfs:
> oh ok I wasn't aware of such a use case. I think supporting both signatures is a bit too complex.

Hey, don't worry, we could just go with 

```lua
---@param switchbuf string|fun(bufnr: integer, line: integer, column: integer):integer?
```

WDYT?

PR #1562, opened 2025-10-04
fall back to `step_into` if `askForTargets` is specified but no targets are available

### LINKED ISSUE #1561: Make `step_into({ askForTarget = true })` fall back to normal `step_into()` if no targets are available

Example for rust, but also happens in C++ if there is nothing to step into (and maybe other times aswell).

```lua
      dap.adapters.codelldb = {
        type = "executable",
        command = "codelldb",
      }

      dap.configurations.rust = {
        {
          -- If you get an "Operation not permitted" error using this, try disabling YAMA:
          --  echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
          name = "Attach to process (codelldb)",
          type = 'codelldb',
          request = 'attach',
          pid = require('dap.utils').pick_process,
          args = {},
        },
      }

```

### Debug adapter version

1.11.5

### Steps to Reproduce

1. Start a debugging session (in my case with codelldb)
2. Use `require('dap').step_into({ askForTarget = true })` to step through your program.
3. You can't step though your program this way, since if there are no targets provided by the debug adapter, `step_into` does nothing in this case.

See this video for a demonstration, where normal `step_into()` steps into a function, but `step_into({ askForTarget = true })` gets no targets from the debug adapter and thus, does not:

https://github.com/user-attachments/assets/781f3bb1-865c-49e1-86ba-3bb363bae471

### Expected Result

That when the debug adapter does not provide any targets, the fallback is `step_into()`, which would then fall back to `step_over()`.

### Actual Result

If the debug adapter does not provide any targets, the message "No target selected. Not stepping" is printed.

Gaweringo:
fixes #1561

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -726,6 +726,12 @@ function M.step_into(opts)
       return
     end
 
+    -- Fall back to step-into/over functionality if there are no targets
+    if #response.targets == 0 then
+      M.step_into(opts)
+      return
+    end
+
     lazy.ui.pick_if_many(
       response.targets,
       "Step into which function?",


mfussenegger:
See https://github.com/mfussenegger/nvim-dap/pull/1562 and https://github.com/mfussenegger/nvim-dap/issues/1561#issuecomment-3445905593. Need to rule out first if this is a unexpected behavior in the debug-adapter

PR #1563, opened 2025-10-12
feature: allow cwd for external terminal

Andy-Zi:
This change allows users to pass the working directory (cwd) from the debug adapter to the external terminal.

Some terminal emulators require an explicit flag to set the working directory. This was not previously possible, as the cwd from the debug adapter was not passed to the external terminal command.

This PR introduces a ${cwd} placeholder that can be used in the external_terminal.args configuration. nvim-dap will substitute this placeholder with the cwd provided by the debug adapter in the runInTerminal request.

### DIFF
--- a/doc/dap.txt
+++ b/doc/dap.txt
@@ -661,28 +661,6 @@ should be launched by nvim-dap:
   }
 <
 
-You can include the placeholder `${cwd}` in `external_terminal.args` to pass
-the working directory requested by the debug adapter (for example from
-`launch.json`). This is useful for terminals that require an explicit working
-directory flag.
-
-Examples:
-
->lua
-  local dap = require('dap')
-  -- Kitty: open in the requested cwd and run the command, keep the tab open
-  dap.defaults.fallback.external_terminal = {
-    command = 'kitty';
-    args = { '--hold', '-d', '${cwd}' };
-  }
-
-  -- alacritty: set working directory explicitly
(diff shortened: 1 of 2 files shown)


mfussenegger:
> Some terminal emulators require an explicit flag to set the working directory. 

This doesn't seem to be the case? They allow changing the working directory, but they default to whatever the parent sets.

E.g. the following will have kitty running with `/tmp` as cwd.

```lua
local uv = vim.uv
local opts = {
  cwd = "/tmp",
  detached = true,
}
local handle, pid_or_err = uv.spawn("kitty", opts, function(code)
end)
```


Or:

```lua
local opts = {
  cwd = "/tmp",
  detached = true,
  args = {
    "--hold", "python", "-c", "import os; print(os.path.abspath(os.curdir))"
  }
}
local handle, pid_or_err = uv.spawn("kitty", opts, function(code)
end)
```

Will print `/tmp`.


I think the whole `${cmd}` substitution can be removed, all that's required is to set the `cwd` property for the uv spawn opts.

Andy-Zi:
Thank you for your feedback, I changed and tested it and it works without the substitution logic.

mfussenegger:
Thanks. Merged via https://codeberg.org/mfussenegger/nvim-dap/commit/6782b097af2417a4c3e33849b0a26ae2188bd7ea

PR #1566, opened 2025-10-23
[feature] Added silent parameter to the dap.repl.execute function to not repeat the command after execution.

banana59:
**Example for Python**
```lua
local repl = require('dap.repl')

repl.commands = vim.tbl_extend('force', repl.commands, {
    custom_commands = {
        ["."] = function(expr)
            if not expr or expr:match("^%s*$") then
                repl.append('Usage: . <expression>')
                return
            end

            local wrapped = 'repr(' .. expr .. ')'
            repl.execute(wrapped, {context='repl', silent=true})
        end,
    }
})
```

Execution in the repl:

```
. <variable>
```
to show a variable without meta information.

I have tested this in python for the nvim-dap commit 6782b09.

This is related to these issues:
- https://github.com/mfussenegger/nvim-dap/issues/737
- https://github.com/mfussenegger/nvim-dap/issues/1062

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -173,6 +173,9 @@ M.defaults = setmetatable(
 
       ---@type nil|fun(session: dap.Session, output: dap.OutputEvent)
       on_output = nil,
+
+      ---@type table<string, boolean> File types where REPL should not show tree structure/metadata
+      repl_hide_tree_filetypes = {},
     },
   },
   {
--- a/lua/dap/repl.lua
+++ b/lua/dap/repl.lua
@@ -205,24 +205,31 @@ local function evaluate_handler(err, resp)
     M.append(tostring(err), nil, { newline = true })
     return
   end
-  local layer = ui.layer(repl.buf)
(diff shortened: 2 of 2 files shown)


mfussenegger:
Isn't this a bit confusing if there's no separation/variable indication in the output?

banana59:
Certainly in some cases, that's why the default is still verbose. I quite liked the neatness in this usecase, that `. dataFrame` directly outputs the content instead of first outputting the interpreted version like `repr(dataFrame)`.

banana59:
I have adjusted the implementation such that now simply entering the variable within the configuration below allows to suppress unwanted metainfo / tree output in repl.

```lua
dap.defaults.fallback.repl_hide_tree_filetypes = {
    python = true,
    python3 = true,
}
```
One could still use the *DAP Scopes* to inspect variables in more detail or for e.g. python one could use `dir(<variable>)`.

banana59:
@mfussenegger Could you please review this PR again?

Issue #1575, opened 2025-12-29
Restart request doesn't wrap config in 'arguments' field (affects Delve rebuild)


fredrikaverpil:
## Problem

It seems like the `restart` request in nvim-dap sends the configuration directly as the request body, but some debug adapters (notably Delve/dlv for Go) expect the configuration to be wrapped in an `arguments` field with additional parameters like `rebuild`.

When using Delve for Go debugging, the restart request completes successfully (showing "Restarted debug adapter"), but the debuggee doesn't actually restart properly because Delve doesn't receive the expected request format.

## Current behavior

In [`lua/dap.lua` line 952](https://github.com/mfussenegger/nvim-dap/blob/master/lua/dap.lua#L952), the restart request sends the config directly:

```lua
lsession:request('restart', config, function(err0, _)
```

## Expected behavior

Delve expects the restart request to be formatted as (see [Delve's test](https://github.com/go-delve/delve/blob/af348314db8506470f6fe5ee51be5f81b20f19ea/service/dap/server_test.go#L6020-L6030)):

```lua
{
  arguments = {
    request = "launch",
    mode = "test",
    program = "...",
    rebuild = true,  -- triggers recompilation before restart
    -- ... other config fields
  }
}
```

The `rebuild` parameter was added in [go-delve/delve#4103](https://github.com/go-delve/delve/pull/4103) to fix restart functionality. It defaults to `true` when present, but it seems like without the proper `arguments` wrapper, Delve may not process the request correctly.

## Possible solution

It seems like the restart request could wrap the config in an `arguments` field:

```lua
lsession:request('restart', { arguments = config }, function(err0, _)
```

Or perhaps provide a way for adapters/users to customize the restart request body format.

## Workaround

Currently, we work around this by disabling `supportsRestartRequest` in the session capabilities, forcing nvim-dap to use terminate+rerun instead:

```lua
require("dap").listeners.after.event_initialized["my-plugin"] = function(session)
  session.capabilities.supportsRestartRequest = false
end
```

## References

- Delve restart issue: https://github.com/go-delve/delve/issues/4102
- Delve fix (adds rebuild param): https://github.com/go-delve/delve/pull/4103
- VS Code Go had the same issue: https://github.com/golang/vscode-go/issues/3835
- DAP spec for restart request: https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Restart

## Environment

- nvim-dap: latest
- Delve: v1.25.2
[...]

PR #1576, opened 2025-12-29
fix: wrap restart config in arguments field per DAP spec

fredrikaverpil:
## Problem

The `restart` request sends the configuration directly as the request body, but according to the [DAP specification](https://microsoft.github.io/debug-adapter-protocol/specification), it should be wrapped in an `arguments` field:

```typescript
interface RestartArguments {
  /**
   * The latest version of the `launch` or `attach` configuration.
   */
  arguments?: LaunchRequestArguments | AttachRequestArguments;
}
```

This causes issues with debug adapters like Delve (Go debugger) that expect the spec-compliant format. When restarting a debug session, the restart request completes successfully but the debuggee doesn't actually restart.

## Solution

Wrap the config in `{ arguments = config }` when sending the restart request, matching the DAP specification.

## References

- Issue: #1575
- DAP specification: https://microsoft.github.io/debug-adapter-protocol/specification (search for `RestartArguments`)
- Delve restart issue: https://github.com/go-delve/delve/issues/4102

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -949,7 +949,7 @@ function M.restart(config, opts)
   if lsession.capabilities.supportsRestartRequest then
     lazy.async.run(function()
       config = prepare_config(config)
-      lsession:request('restart', config, function(err0, _)
+      lsession:request('restart', { arguments = config }, function(err0, _)
         if err0 then
           notify('Error restarting debug adapter: ' .. tostring(err0), vim.log.levels.ERROR)
         else


fredrikaverpil:
To clarify the issue - there's confusingly nested `arguments` in the DAP protocol:

**Current nvim-dap sends:**
```json
{
  "command": "restart",
  "arguments": { "type": "go", "request": "launch", "program": "..." }
}
```

**DAP spec expects:**
```json
{
  "command": "restart", 
  "arguments": {
    "arguments": { "type": "go", "request": "launch", "program": "..." }
  }
}
```

The outer `arguments` is from the base DAP `Request` interface, the inner `arguments` is from `RestartArguments`. It's confusing naming in the spec, but that's what it requires.

Delve specifically parses `request.Arguments.Arguments` in their [restart handler](https://github.com/go-delve/delve/blob/af348314db8506470f6fe5ee51be5f81b20f19ea/service/dap/server.go#L3096-L3106), which is why the current format doesn't work. Other adapters might be more lenient or ignore restart arguments entirely.

mfussenegger:
Thanks

PR #1580, opened 2026-01-08
feat: add dap.pick_process listener

jinzhongjia:
This will handler vscode launch json configuration like this:

```json
{
      "name": "Attach to Process",
      "type": "go",
      "request": "attach",
      "mode": "local",
      "processId": 0
},
```

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -422,6 +422,15 @@ do
   M.listeners.on_config["dap.expand_variable"] = function(config)
     return vim.tbl_map(expand_config_variables, config)
   end
+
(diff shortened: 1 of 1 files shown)


jinzhongjia:
now, processId 0 will automatically pick process

mfussenegger:
Thanks, but a `processId` property is not part of the debug adapter protocol specification but specific to concrete adapter implementations like cppdbg. gdb and codelldb for example instead use `pid`.

Because of that, this is out of scope of nvim-dap core. Given that the listeners are pluggable this can be added to one's configuration or would fit into a extension centered around a specific adapter.

PR #1582, opened 2026-01-10
refactor(dap): ensure run_last fetches the latest configuration by name

kwdiwt:
The current implementation of run_last caches the configuration object from the previous session. This leads to "stale" runs if the user modifies their launch.json or if a dynamic configuration provider updates its values (e.g., changing an environment variable from "RUN_MODE": "a" to "b").

Example Case: When debugging a Python module with a launch.json like this:
```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "main",
      "type": "debugpy",
      "request": "launch",
      "module": "mobile.main",
      "env": {
        "RUN_MODE": "a",
      }
   }
}
```
If the user changes "RUN_MODE" to "b", the updated run_last will correctly pick up the new environment variable without requiring the user to re-select the configuration manually.

### DIFF
--- a/lua/dap.lua
+++ b/lua/dap.lua
@@ -668,13 +668,27 @@ function M.run(config, opts)
 end
 
 
---- Run the last debug session again
+--- Re-fetches the configuration by name from providers and executes it.
+--- This is useful when the environment or dynamic configuration changes
+--- between runs, as it ensures the latest config values are used.
 function M.run_last()
-  if last_run then
-    M.run(last_run.config, last_run.opts)
-  else
-    notify('No configuration available to re-run', vim.log.levels.INFO)
+  if not last_run then
+    return
   end
+
+  local bufnr = api.nvim_get_current_buf()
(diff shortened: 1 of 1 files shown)


kwdiwt:
or add M.run_last_by_name ?

mfussenegger:
> or add M.run_last_by_name ?

I'm not to keen on adding different versions of the function. If so I'd rather tend to a `opts` parameter that adds options to control the behavior. There have also been requests for showing a list of recent configurations, or to avoid re-evaluation of function properties.

But before that I'd like to see more interest in this.

In the meantime you can as workaround also add a custom version in your dotfiles with something like:

```lua
local dap = require("dap")
local last_config = nil

---@param session Session
dap.listeners.after.event_initialized["store_config"] = function(session)
  last_config = session.config
end

function M.run_last()
  if last_config then
    -- resolve config like in this PR
  else
    dap.continue()
  end
end
```

PR #1592, opened 2026-02-22
send empty array for setbreakpoints command

Jaehaks:
1. Problem

Some debugger like matlab-lsp, It removes unused breakpoints when lsp received setbreakpoints command.

Previously, set_breakpoints() function send nothing when there are no breakpoint. It makes that lsp cannot remove remained breakpoints which are set before.

2. Solution

Send setbreakpoints request with empty array even though there is no breakpoints

### DIFF
--- a/lua/dap/session.lua
+++ b/lua/dap/session.lua
@@ -953,10 +953,8 @@ do
   function Session:set_breakpoints(bps, on_done)
     local num_requests = vim.tbl_count(bps)
     if num_requests == 0 then
-      if on_done then
-        on_done()
-      end
-      return
+	  local bufnr = vim.api.nvim_get_current_buf()
+      bps = { [bufnr] = {} }
     end
     for bufnr, buf_bps in pairs(bps) do
       notify_if_missing_capability(buf_bps, self.capabilities)


mfussenegger:
Can you clarify how you run into this?

If I remove the last breakpoint in a buffer I still get a `setBreakpoints` command sent like:

```
{
  arguments = {
    breakpoints = {},
    lines = {},
    source = {
      name = "Foo.java",
      path = "/tmp/Foo.java"
    },
    sourceModified = false
  },
  command = "setBreakpoints",
  seq = 11,
  type = "request"
}
```

Jaehaks:
When I executes `continue()` function these messages are sent from nvimd-dap If I didn't check any breakpoints.

There is no setbreakpoint message between `initialize` and `launch` 

```lua
Content-Length: 343

{"type":"request","arguments":{"linesStartAt1":true,"columnsStartAt1":true,"pathFormat":"path","clientName":"neovim","locale
":"en_US","adapterID":"nvim-dap","clientID":"neovim","supportsStartDebuggingRequest":true,"supportsProgressReporting":true,"
supportsVariableType":true,"supportsRunInTerminalRequest":true},"command":"initialize","seq":1}
Content-Length: 212

{"type":"request","arguments":{"program":"C:\\Users\\USER\\.config\\nvim\\test\\test.m","type":"matlab","cwd":"C:\\Users\\US
ER\\.config\\nvim","request":"launch","name":"MATLAB Debug"},"command":"launch","seq":2}
Content-Length: 56

{"type":"request","command":"configurationDone","seq":3}
Content-Length: 145

{"type":"request","arguments":{"expression":"run('C:\\Users\\USER\\.config\\nvim\\test\\test.m')","context":"repl"},"command
":"evaluate","seq":4}
```

I created a dap adapter for matlab. To communicate between  dap and lsp, I made some `read_handler()` which send dap message to lsp when the tcp socket received message from dap, 
You can find this in [nvim-dap-matlab](https://github.com/Jaehaks/nvim-dap-matlab/blob/main/lua/nvim-dap-matlab/adapter.lua).

When I modify your code like PR, It send empty message correctly.

mfussenegger:
>  If I didn't check any breakpoints. There is no setbreakpoint message between initialize and launch

That's sort of expected. `setBreakpoints` is optional - a client doesn't have to send it if there are no breakpoints.

> It makes that lsp cannot remove remained breakpoints which are set before.

If the language server keeps the breakpoints, couldn't you just always remove breakpoints at the end of a debug session?

